PR #5, opened 2014-12-30
Basic framework for QuArrays and basis types

jrevels:
This PR splits off the non-QuDirac work from #4. To copy the summary from there, here's what this PR does:
1. Provides a cleaner directory structure and a more comprehensive type hierarchy.
2. Organizes/cleans up existing code, as well as splits off QuDirac-related work into the [QuDirac repo, which got moved under the JuliaQuantum organization.](https://github.com/JuliaQuantum/QuDirac.jl) That repo is where abstract Dirac-notated states and operator functionality will live for the time being, whereas QuBase will be focused on: 
   - solvers/modeling capability for coefficient arrays
   - a foundational type system
   - a solid, general implementation for `QuArray`s and fundamental basis types
3. Updates the `creationop` method provided by @acroy to work with arbitrary number of particles (you can pick which particle the operator should act on). I renamed it to `raiseop` for brevity's sake and implemented `lowerop` as well. These methods aren't optimized in the slightest (likely a tad slower than @acroy's original implementation), but provide a good starting point for ladder operations in multi-particle systems.
4. Gives a single type implementation of `FiniteBasis` that supports a tensor product structure
5. Implements a couple of functions for `QuArray`/`FiniteBasis`, including pretty-printing for `QuArray`s (let me know what you think, it's hard to compactly convey all that info).

Sorry for the previous commit; I had to go back and fix the tabs-to-spaces thing again for some reason.


.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+    - JULIAVERSION="juliareleases"
+    - JULIAVERSION="julianightlies"
+before_install:
+  - sudo add-apt-repository ppa:staticfloat/julia-deps -y
+  - sudo add-apt-repository ppa:staticfloat/${JULIAVERSION} -y
+  - sudo apt-get update -qq -y
+  - sudo apt-get install libpcre3-dev julia -y
+  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+script:
+  - julia --check-bounds=yes -e 'versioninfo(); Pkg.init(); Pkg.clone(pwd()); Pkg.build("QuBase"); Pkg.test("QuBase")'
```
acroy: We can use the [simplified version of the .yml file](https://groups.google.com/forum/?nomobile=true#!searchin/julia-users/Travis/julia-users/BtCxh4k9hZA/ngUvxdxOxQ8J) now.

src/arrays/ladderops.jl
```diff
@@ -0,0 +1,68 @@
...
+            end
+            arr = kron(gen_op_mat(lens[n], ladderflag), arr)
+        end 
+        for i=n+1:length(lens)
+            arr = kron(speye(Complex128, lens[i]), arr)
+        end
+        return arr
+    end
+
+export raisematrix,
```
acroy: Do we need/want to export `raisematrix` and `lowermatrix`?

src/arrays/quarray.jl
```diff
@@ -0,0 +1,146 @@
...
+    getindex(qa::QuArray, i...) = getindex(qa.coeffs, i...)
+
+    in(c, qa::QuArray) = in(c, qa.coeffs)
+
+    ctranspose(qa::QuVector) = QuArray(ctranspose(qa.coeffs), qa.bases)
+    ctranspose(qa::QuMatrix) = QuArray(ctranspose(qa.coeffs), reverse(qa.bases))
+    transpose(qa::QuVector) = QuArray(transpose(qa.coeffs), qa.bases)
+    transpose(qa::QuMatrix) = QuArray(transpose(qa.coeffs), reverse(qa.bases))
+
+    zeros(qa::QuArray) = QuArray(zeros(qa.coeffs), qa.bases)
```
acroy: Could you add `eye(qa::QuArray)` as well?

src/bases/finitebasis.jl
```diff
@@ -0,0 +1,60 @@
...
+    #
+    # F_4 x F_3 x F_2 => {|0,0,0>, |0,0,1>, |0,1,0>, |0,1,1>...|3,2,1>}
+    # 
+    # Note that the subscript number refers to basis *length*, not max
+    # excitation level.
+    #
+    # This tensor product basis, F_4 x F_3 x F_2, could be represented 
+    # by calling FiniteBasis(4, 3, 2). 
+
+    immutable FiniteBasis{S} <: AbstractBasis{S}
```
acroy: Would it be advantageous to use `NTuple` here? I don't have any experience with NTuples vs Tuples, but it appears to be used in Base for similar purposes (for example  [DArrays](https://github.com/JuliaLang/julia/blob/12ce837cd5774a41e94c7b329e75213c5b84af25/base/darray.jl)).

.travis.yml
jrevels: Awesome catch! I hadn't seen that this had happened, we should take advantage of it for sure.

src/arrays/ladderops.jl
jrevels: That's a good point - we're surely going to end up implement a lot of operations that act perfectly validly on general `AbstractArray` types outside of `QuArray`. Do we want to provide that functionality to the world? I don't see any reason why not, it only makes the package more generally useful, and - at least in this case - doesn't expose anything obviously harmful.

src/arrays/quarray.jl
jrevels: Done!

src/bases/finitebasis.jl
jrevels: You mean instead of `lens::(Int...)` we'd have `lens::NTuple{N,T}`? I'd like to be able to support different lengths for the factor bases - `NTuple` would work for multiple factor bases of uniform length, though, if we wanted to make a separate basis type for that case.

src/arrays/ladderops.jl
acroy: Personally, I think we should export as few things as possible - at least in the beginning. And in particular I would try not to export abstract types. Maybe we should have a separate issue on export policy?

src/bases/finitebasis.jl
acroy: Yes, I wanted to replace `lens::(Int...)` by `lens::NTuple{N,Int}`, but you are right, in that case `N` needs to be a type parameter of `FiniteBasis`. Would that be problematic? I am just "concerned" that the performance of Tuples is limiting us at some point?

src/arrays/ladderops.jl
```diff
@@ -0,0 +1,68 @@
...
+    ##########################
+    # Helper Functions/Types #
+    ##########################
+    abstract LadderOpFlag
+    abstract RaiseOpFlag <: LadderOpFlag
+    abstract LowerOpFlag <: LadderOpFlag    
+
+    ladder_inds(n, ::Type{RaiseOpFlag}) = ([1:n-1], [2:n])
+    ladder_inds(n, ::Type{LowerOpFlag}) = ([2:n], [1:n-1])
+    laddercoeffs(n) = complex(sqrt(linspace(1, n, n)))
```
acroy: We don't need `complex` here (might be a leftover from my code).

src/bases/finitebasis.jl
jrevels: I can't think of any significant performance issues with using a tuple type here - what did you have in mind?

Unless we are explicitly using `NTuple` as a field, or some similarly typed field, I'm weary of making `N` a parameter of `FiniteBasis` for the type stability issues I've previously discussed. Are you using `N` as the number of states, or the number of factor bases? The former I'm against, because that would be an abuse of the type system/probably lead to type stability issues - it would be like storing `length` as a type parameter for `Array`. I'm more open-minded but still reluctant about the latter.

src/arrays/ladderops.jl
jrevels: I think it merits it does indeed merit its own issue. I don't have time tonight to write one up but if you would like to, I'd be appreciative. If not, I can post one sometime this week.

src/bases/finitebasis.jl
acroy: I actually meant the latter (if `N=length(lens)` is the number of factor basis). I have to say that I don't know if there are performance issues with Tuples. It's just that I saw `NTuples` frequently used in such situations and I thought I should bring it up now.

src/arrays/ladderops.jl
jrevels: It's actually slightly more performant to do the conversion here than at the time of `setindex!` when filling the array in `fill_op_arr!` (this is of course assuming that we want our "default" element type to be `Complex128`, which I believe we do). I think the performance boost is due to it being easier for the compiler to optimize a map versus the mutation that would otherwise have to occur in the `for` loop, though I'm not entirely sure.

src/bases/finitebasis.jl
jrevels: After playing around with `NTuple`s more, I see I was off-base in what my assumptions were about how they worked. This is actually great, I've never messed with these before but you can be sure they're going to see a lot of use over in `QuDirac` as well...

My comment before about `NTuple`s not being suitable for non-uniform factor bases is invalid - I'll make the change this weekend. 

Also, I don't think there is anything to worry about in regards to tuple performance - they're more compact than normal arrays and their immutability gives the compiler a slight advantage in some cases (of course, there are many instances where mutable arrays *should* be used, but I think that `NTuple` really is the right way to go here).

src/arrays/ladderops.jl
acroy: Sorry, my comment wasn't really clear. I meant to change the element type to `Float64`. I think we can rely on the "default" element types and promotion rules of the coefficient container.

src/bases/finitebasis.jl
acroy: Great! I really have to look into `NTuples` as well. Have you found any documentation for them?

jrevels: I haven't, but it seems that it pulls some tricks to allow you to encode the length of the tuple in the type. This theoretically makes sense, since a tuple type in and of itself can be a data type, but in practical use it's hard to dispatch on, since the compiler often treats tuple types as values when you're acting on them. For example:
```
julia> type Foo{N}
           tup::Tuple
       end

julia> Foo{T<:Tuple}(tup::T) = Foo{length(T)}(tup)

julia> [Foo((1,2,3)) for i=1:2]
2-element Array{Foo{N},1}:
 Foo{3}((1,2,3))
 Foo{3}((1,2,3))
```
While the outer constructor above is theoretically type stable, Julia's compiler seems to have trouble inferencing types constructed in this manner - as you can see, it couldn't guess the proper element type for that array. I believe this is because `T<:Tuple` is considered to be a "value"  rather than a "type" when consumed by the `length` function. With `NTuple`s, however, one can reliably encode this information:

```
julia> type Bar{N}
           tup::NTuple{N}
       end

julia> [Bar((1,2,3)) for i=1:2]
2-element Array{Bar{3},1}:
 Bar{3}((1,2,3))
 Bar{3}((1,2,3))
```

.travis.yml
jrevels: Done!

src/bases/finitebasis.jl
jrevels: Made the switch to `NTuple`s for `FiniteBasis`!

src/arrays/ladderops.jl
jrevels: Done! Removed the parts of the code that was forcing `Complex128` by default.

src/arrays/quarray.jl
```diff
@@ -0,0 +1,148 @@
...
+        bases::B
+        function QuArray(coeffs::AbstractArray{T, N}, bases::B) 
+            if checkbases(coeffs, bases) 
+                new(coeffs, bases)
+            else 
+                error("Coefficient array does not conform to input bases")
+            end
+        end
+    end
+
```
jrevels: So now that we know about `NTuple`s, how do folks feel about changing `QuArray` to this:

```
abstract AbstractQuArray{B<:AbstractBasis, T, N} <: AbstractArray{T,N}

type QuArray{B<:AbstractBasis, T, N, A} <: AbstractQuArray{B, T, N}
    coeffs::A
    bases::NTuple{N, B}
    function QuArray(coeffs::AbstractArray{T}, bases::NTuple{N, B}) 
        if checkbases(coeffs, bases) 
            new(coeffs, bases)
        else 
            error("Coefficient array does not conform to input bases")
        end
    end
end
```

I believe the above to be much, much cleaner. There are two significant changes with this approach:

1. There can only be a single basis type for the entire Array. I don't really see any detriment to 
this - when would you ever want different bases types along different dimensions? If anything, the 
restriction makes the whole object easier to reason about.

2. We define the number of dimensions of a `QuArray` to be equivalent to how many bases its storing. 
This way, `QuArray{..., 1}` really does act like a vector, `QuArray{..., 2}` really does act like an
operator, and `QuArray{..., 3+}` really does act like a tensor. This approach allows us to sidestep 
some of the issues with the conflation of vectors and matrices. I was already heading down this road
with the `QuVector` and `QuMatrix` type aliases, but this is much more clearly laid out.

acroy: Nice! I guess I would have asked you about `NTuple`s and `QuArray`s anyways :-) I also think that this change makes the definition of `QuArray`s much cleaner and nicer. The only objection I can think of is your point 1), because it "restricts" us to cartesian tensors only (ie we don't distinguish between raised and lower indices anymore). However, I would say that this is a reasonable compromise, since it will be the most common use case(?).

jrevels: I agree. I've implemented this now, let me know what you think. If there aren't any other issues (besides exportation, which we now have a separate issue for), I'm ready for this to be actually merged.

src/arrays/quarray.jl
```diff
@@ -0,0 +1,143 @@
...
+    end 
+    getindex(qa::QuArray, i::AbstractArray) = getindex(qa.coeffs, i)
+    getindex(qa::QuArray, i::Real) = getindex(qa.coeffs, i)
+    getindex(qa::QuArray, i) = getindex(qa.coeffs, i)
+    getindex(qa::QuArray, i...) = getindex(qa.coeffs, i...)
+
+    in(c, qa::QuArray) = in(c, qa.coeffs)
+
+    ctranspose(qa::QuVector) = QuArray(ctranspose(qa.coeffs), qa.bases)
+    ctranspose(qa::QuMatrix) = QuArray(ctranspose(qa.coeffs), reverse(qa.bases))
```
acroy: I like the fact that a transposed `QuVector` stays a `QuVector` and doesn't become a matrix like for arrays. But we might want to have some way to indicate that a `QuVector` or a `QuMatrix` was transposed. I am thinking of something like JuliaLang/julia#6837, i.e. introducing a dual/transposed type. This would make it easier to define inner products and such. We don't have to do this now. Maybe once this is merged we can have an issue as a reminder. cc: @Jutho

src/bases/finitebasis.jl
```diff
@@ -0,0 +1,61 @@
+import Base: length,
+    size
```
acroy: Might need to import `show` and `convert` as well.

acroy:
I think this looks good. Could you maybe squash some of the commits (or all of them)?

src/bases/finitebasis.jl
jrevels: Done!

src/arrays/quarray.jl
jrevels: Yeah, the transpose distinction is going to need to exist at some point. I have a `DualType` for DiracArrays in QuDirac, but the way I'm using it doesn't extend to anything beyond vectors and I'd rather have the distinction happen in QuBase anyway (or even better, in Julia base, but who knows how long that will take). It's a tough nut to crack. I think giving it an issue is the right way to go (hopefully one that links to all the relevant issues elsewhere so we don't repeat discussion).

PR #11, opened 2015-02-05
Eager QuCoeffs

jrevels:
This PR is competing with #10 as a proposal for a coefficient container type for QuArrays. 

There are few key advantages:
1. Conjugation/transposition is performed eagerly. This is closer to standard Julia Array behavior than implementing a lazy transpose, and should mean less work for us moving forward, since more functions will be able to be simply delegated to the `arr` field without having to worry about moving around indices. 
2. Now that everything is eagerly performed anyway, we have no need for disambiguating a "dual" vector from the conjugate transposed vector. This is much cleaner than having a `dual` function and a `ctranspose` function, IMO.
3. Note that transposition for an N=1 instance of QuCoeffs is simply a copy with the transpose flag flipped. This allows us to represent row vectors as single dimensional.

Let me know what you think. Personally, I greatly prefer this approach to that of #10.

### DIFF
--- a/src/QuBase.jl
+++ b/src/QuBase.jl
@@ -54,7 +54,7 @@ module QuBase
     # Include Statements #
     ######################
         include("bases/bases.jl")
-        include("arrays/quarray.jl")
+        include("arrays/arrays.jl")
     
     export AbstractStructure, 
         AbstractQuantum,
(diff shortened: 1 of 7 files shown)


jrevels:
No negative response on this in 16 days, I'm gonna merge it in.

PR #12, opened 2015-02-27
Correct typos in typealias for QuVector and QuMatrix.

acroy:


jrevels:
Ugh, I gotta get better about these careless mistakes. Thanks.

PR #13, opened 2015-02-27
Correct coefficients for raising and lowering operators.

acroy:
The coefficients of the raising and lowering operators were swapped.


jrevels:
So it is, sorry about that. Nice catch.

PR #14, opened 2015-03-01
Simpler CTranspose + multiplication methods

jrevels:
Okay, hopefully this will be the last attempt needed to kill #9 so we can finally move forward!

This PR defines bunch of multiplication methods that are consistent, using @acroy's method to resolve the row/column vector issues. It also adds actual tests for these methods. They're not exactly comprehensive, but it's a start.

Actually defining the multiplication methods made me realize how messy having separate Conjugate/Transpose types makes things, so I simply removed those types from the picture. If users wish to separately conjugate/transpose coefficient arrays, we'll let them worry about that (or just do so eagerly).

Thoughts?


jrevels:
Also it should be noted that, in the future, we could set up type promotion for basis types that would allow us to expand beyond forcing both arguments to have the same basis type.

src/arrays/arraymath.jl
```diff
@@ -0,0 +1,55 @@
+import Base: *
+
+##################
+# Multiplication #
+##################
+# bra * ket -> scalar
+function *{B}(bra::DualVector{B}, ket::QuVector{B})
+    return first(Ac_mul_B(rawcoeffs(bra), rawcoeffs(ket)))
```
acroy: Should we use `dot` here?

acroy:
Thanks for putting this together. Looks good at a first glance. I will try it out tomorrow. You forgot to call `runtests.jl` in `travis.yml`. We probably have to activate Travis for this repo as well.

acroy:
Ok, Travis should work with the next push/PR (forget the comment about `runtests.jl`). I have also added a status banner to README.

Otherwise the PR looks good. There are just two things we should keep in mind (and maybe put a comment in the code): 1) Currently we ignore the inner product of the basis functions, i.e., assume an orthonormal basis. 2) The multiplication doesn't work for `N>2`. We probably want to extend this at least to `N=4` (super-operators), but ideally want it to work in general (or at least make it easy to extend).

src/arrays/arraymath.jl
jrevels: Done.

acroy:
I don't know why Travis complains...

jrevels:
Did it complain? Everything's turning up green as far as I can see, but I'm not super familiar with Travis so I'm probably missing something. Anyway, thanks for setting it up. 

Agreed on the generality points you made. I think for basis products we're going to want to define the functions `inner{A,B,S}(a::NTuple{A,FiniteBasis{S}}, b::NTuple{B,FiniteBasis{S}})` and
`tensor{A,B,S}(a::NTuple{A,FiniteBasis{S}}, b::NTuple{B,FiniteBasis{S}})`, then use those to construct more general multiplication definitions. 

Just to be on the safe side, I'll restrict the functions defined here to S<:Orthonormal in the meantime.

acroy:
Now it looks good (I.e. green). It was probably just the old yml file.

PR #15, opened 2015-03-04
Added Normalize to QuArray.

amitjamadagni:
This is my first attempt at contributing. Though not that significant, please do comment. Thanks.

### DIFF
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -1,12 +1,12 @@
-import Base: *
+import Base: *, norm, scale, scale!
 
 ##################
 # Multiplication #
 ##################
 
 # The below needs to be defined more generically;
-# the below only defines multiplication on orthonormal, 
-# N<3 objects 
+# the below only defines multiplication on orthonormal,
+# N<3 objects
 
 typealias OrthonormalBasis{S<:Orthonormal} AbstractBasis{S}
 
@@ -16,16 +16,16 @@ function *{B<:OrthonormalBasis}(bra::DualVector{B}, ket::QuVector{B})
 end
 
 # bra * operator -> bra
-function *{B<:OrthonormalBasis}(bra::DualVector{B}, op::QuMatrix{B}) 
+function *{B<:OrthonormalBasis}(bra::DualVector{B}, op::QuMatrix{B})
     return QuArray(Ac_mul_B(rawcoeffs(op), rawcoeffs(bra)), bases(op,2))'
 end
 
-function *{B<:OrthonormalBasis}(bra::DualVector{B}, op::DualMatrix{B}) 
+function *{B<:OrthonormalBasis}(bra::DualVector{B}, op::DualMatrix{B})
     return QuArray(rawcoeffs(op)*rawcoeffs(bra), bases(op,2))'
 end
 
 # operator * ket -> ket
-function *{B<:OrthonormalBasis}(op::QuMatrix{B}, ket::QuVector{B}) 
+function *{B<:OrthonormalBasis}(op::QuMatrix{B}, ket::QuVector{B})
     return QuArray(rawcoeffs(op)*rawcoeffs(ket), bases(op,1))
 end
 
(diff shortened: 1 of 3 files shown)


acroy:
Thanks for your contribution! It's a good start. Accidentally I thought that adding norms would be a natural  next step ...

Regarding your code I have some comments (mostly because we don't have an explicit guide how such functions for QuArrays should work, but I have many implicit ideas ;-)
- Firstly, norms are not so easy. You used `Base.norm` on the container, which computes the "p-norm". However, for (Hilbert space) operators it is more common to use the [Frobenius norm](http://en.wikipedia.org/wiki/Matrix_norm), which you would get via `Base.vecnorm`. For vectors this doesn't matter, but for `N>1` we should probably use the latter. Note that we need a different norm (i.e. `trace`) for density matrices though.
- I would prefer to have a separate function `norm(qa::QuArray)` (and probably also for `CTranspose`).
- For scaling we should provide a mutating function by extending `Base.scale!` and then implement multiplication and division in terms of that. You will need a `copy` function first (@jrevels Could you add `copy` for `QuArray` and `CTranspose`?).
- Note that your current `normalize!` function actually makes a copy of the coefficient array.
- I think `coeffs` was replaced by `rawcoeffs` in one of the latest commit?
- Now that we have Travis in place, which runs tests for each PR, we should really try to add tests for each function/operation we add.

Sorry for the long list of comments. Maybe the more general points should go into some "guideline issue".

amitjamadagni:
@acroy I have worked on the commit. Here are the details :
* Attempt at implementing the `norm` function which takes `QuArray` as a parameter.
* Attempt at implementing copy for `QuArray` and `CTranspose`.
* `normalize`, `normalize!` implemented.

WIP :
[ ] Scaling 
[ ] Tests 

Doubts :
1. Requirement of rawcoeffs when .coeffs is doing the same job ??

src/arrays/arraymath.jl
```diff
@@ -34,29 +34,53 @@ function *{B<:OrthonormalBasis}(op::DualMatrix{B}, ket::QuVector{B})
...
 *(dm1::DualMatrix, dm2::DualMatrix) = (dm1.qarr*dm2.qarr)'
+
+# scaling
+function *(num::Number, qarr::QuArray)
+    return QuArray(num*qarr.coeffs, qarr.bases)
+end
+
+function norm(qarr::QuArray)
+    if ndims(rawcoeffs(qarr)) == 1
+        return norm(rawcoeffs(qarr))
```
acroy: You can always use `vecnorm`. For `N=1` this is the same as `norm` (sum of absolute values squared of all elements).

src/arrays/arraymath.jl
```diff
@@ -34,29 +34,53 @@ function *{B<:OrthonormalBasis}(op::DualMatrix{B}, ket::QuVector{B})
...
+    end
+end
+
+function normalize(qarr::QuArray)
+    return normalize!(copy(qarr))
+end
+
+function normalize!(qarr::QuArray)
+    for i in 1:length(qarr.coeffs)
+        qarr.coeffs[i] = qarr.coeffs[i]/norm(qarr)
```
acroy: Might be more efficient(?) to use `scale!(norm(qarr), qarr.coeffs)`. For some container types this will for example call a BLAS routine which is more optimized/accurate.

acroy:
Nice! The point of using `rawcoeffs` is that you can use it for QuArrays and CTranspose, i.e., `norm` could take `qa::Union(QuArray,CTranspose)` and you can simply call `rawcoeffs(qa)`.

src/arrays/arraymath.jl
```diff
@@ -34,29 +34,53 @@ function *{B<:OrthonormalBasis}(op::DualMatrix{B}, ket::QuVector{B})
...
 end
 
 *(dm1::DualMatrix, dm2::DualMatrix) = (dm1.qarr*dm2.qarr)'
+
+# scaling
+function *(num::Number, qarr::QuArray)
+    return QuArray(num*qarr.coeffs, qarr.bases)
+end
+
+function norm(qarr::QuArray)
```
acroy: You have to use `function Base.norm` or `import Base: norm` otherwise `norm` will only be visible within the module.

src/arrays/arraymath.jl
amitjamadagni: My bad !!! Sorry for this. Done editing.

src/arrays/quarray.jl
```diff
@@ -34,6 +34,8 @@
     bases(qarr::QuArray, i) = qarr.bases[i]
     bases(qarr::AbstractQuArray) = ntuple(ndims(qarr), i->bases(qarr, i))
 
+    copy{B,T,N,A}(qa::QuArray{B,T,N,A}) = QuArray(copy(qa.coeffs), copy(qa.bases))
```
acroy: Use `Base.copy{B,T,N,A}` instead to extend the function in Base and to make it visible outside QuBase. (I think you don't need the type parameters here, so `qa::QuArray` should be sufficient.

src/arrays/quarray.jl
```diff
@@ -70,6 +72,8 @@
     revind(len, i) = len - (i-1)
     bases(ct::CTranspose, i) = bases(ct.qarr, revind(ndims(ct), i))
 
+    copy{B,T,N,A}(ct::CTranspose{B,T,N,A}) = CTranspose{B,T,N,A}(copy(ct.qarr))
```
acroy: Same as above. Use `Base.copy` and omit type parameters.

acroy:
If you use `rawcoeffs` and `bases` instead of accessing the fields (`qa.coeffs`) you can replace the arguments `qa::QuArray` by `qa::Union(QuArray,CTranspose)` (I think). Don't forget to add `norm` to the import at the top of arraymath.jl. For the test you can basically use the same approach as Jarrett.

src/arrays/quarray.jl
amitjamadagni: @acroy I was just imitating the copy type parameters, I have observed that at some places it is used and others it is not though I could not understand the reason behind this. Any inputs on this would be helpful.

acroy: It's not wrong to use the type parameters, but simply not necessary. `copy{B,T,N,A}=...` will give you a specialized version of `copy` for each combination of `B,T,N,A`. However, since `qa::QuArray` already matches all those types (the type without type parameters is an abstract type; see below for an example), you can omit the type parameters. Roughly speaking, if you don't need the parameters in your function or you don't want to constrain your types (for example in `*` we want the arrays to have the same bases, so we need `*{B}...`), you can most likely omit them.

Regarding the "parametrized types without parameters are abstract" point consider this:
```
type Foo{T<:Number}; n::T; end
julia> isleaftype(Foo)
false

julia> isleaftype(Foo{Int})
true

julia> Foo{Int} <: Foo
true

julia> Foo{Float64} <: Foo
true
```
It's sometimes a bit confusing, but reading the [manual section on parametric types](http://julia.readthedocs.org/en/release-0.3/manual/types/#man-parametric-types) 3-4 times helps :-)

amitjamadagni:
@acroy is this good to go ?? Any comments would be helpful.

src/arrays/arraymath.jl
```diff
@@ -1,12 +1,13 @@
 import Base: *
+import Base: norm, scale, scale!
```
acroy: Those imports you can just add to the line above.

src/arrays/arraymath.jl
```diff
@@ -34,29 +35,57 @@ function *{B<:OrthonormalBasis}(op::DualMatrix{B}, ket::QuVector{B})
...
 *(dm1::DualMatrix, dm2::DualMatrix) = (dm1.qarr*dm2.qarr)'
+*(num::Number, qarr::QuArray) = scale(num, qarr)
+/(qarr::QuArray, num::Number) = scale(1/num, qarr)
+
+function scale(num::Number, qarr::QuArray)
+    return scale!(num, copy(qarr))
+end
+
+function scale!(num::Number, qarr::QuArray)
+    scale!(num, qarr.coeffs)
```
acroy: Why not `rawcoeffs(qarr)` instead of `qarr.coeffs`? As I said earlier, if you use `rawcoeffs` you can change `qarr::QuArray` to `qarr::Union(QuArray,CTranspose)` and your functions will work for both types. Does that not work?

src/arrays/arraymath.jl
```diff
@@ -34,29 +35,57 @@ function *{B<:OrthonormalBasis}(op::DualMatrix{B}, ket::QuVector{B})
...
+function norm(qarr::QuArray)
+    return vecnorm(rawcoeffs(qarr))
+end
+
+function normalize(qarr::QuArray)
+    return normalize!(copy(qarr))
+end
+
+function normalize!(qarr::QuArray)
+    scale!(1/norm(qarr), qarr.coeffs)
```
acroy: Same thing here regarding `rawcoeffs(qarr)` and changing the argument type.

test/multest.jl
```diff
@@ -20,4 +20,10 @@ qv = QuArray(v)
...
 
-@assert qv'*qm*qv == first(v'*m*v)
\ No newline at end of file
+@assert qv'*qm*qv == first(v'*m*v)
+
+@assert rawcoeffs(2*qv) == 2*v
+@assert rawcoeffs(qv/2) == v/2
+@assert rawcoeffs(scale(2,qv)) == 2*v
+@assert norm(qv) == norm(v)
+#@assert rawcoeffs(normalize(qv)) == rawcoeffs(qv)/norm(v)
```
acroy: What is wrong with this test?

amitjamadagni: normalize(qv).coeffs = [0.2672612419124244,0.5345224838248488,0.8017837257372732]
qv.coeffs/norm(v) = [0.2672612419124244,0.5345224838248488,0.8017837257372733]
Hence the test fails.

acroy: You might want to use `@test_approx_eq` or `@test_approx_eq_eps` here (see docs for [Base.Test](http://julia.readthedocs.org/en/release-0.3/stdlib/test/)).

src/arrays/arraymath.jl
amitjamadagni: I had some problems testing it out locally. Now I have tested this out. Sorry I missed on this union type from the previous comments.

acroy:
Looks good now. Thanks! Just one minor point: could you [squash your commits into a single one](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/)? It's not really necessary but makes the history more readable.

amitjamadagni:
@acroy done. Thanks for all the help.

PR #17, opened 2015-03-10
Behavior of ctranspose.

amitjamadagni:
ctranpose has the following behavior :

For 
Input :

m = [1+1im 2+2im 3+3im; 
     4+4im 5+5im 6+6im; 
     7+7im 8+8im 9+9im]
qm = QuArray(m)
qm
3x3 QuMatrix in QuBase.FiniteBasis{QuBase.Orthonormal,1}:
...original coefficients: Array{Complex{Int64},2}
Complex{Int64}[1 + 1im 2 + 2im 3 + 3im
               4 + 4im 5 + 5im 6 + 6im
               7 + 7im 8 + 8im 9 + 9im]

c_qm = ctranspose(qm)
c_qm
3x3 DualMatrix in QuBase.FiniteBasis{QuBase.Orthonormal,1}:
...original coefficients: Array{Complex{Int64},2}
Complex{Int64}[1 + 1im 2 + 2im 3 + 3im
               4 + 4im 5 + 5im 6 + 6im
               7 + 7im 8 + 8im 9 + 9im]

In the above case the conjugate transpose is not performed. 

The pull request aims to fix this. Hope my understanding of the ctranspose is right. Please do comment. Thanks.

### DIFF
--- a/src/arrays/quarray.jl
+++ b/src/arrays/quarray.jl
@@ -91,8 +91,13 @@ import Base: copy
     Base.setindex!(ct::CTranspose, x, i, j) = setindex!(ct.qarr, x', j, i)
     Base.setindex!(dv::DualVector, x, i) = setindex!(dv.qarr, x', i)
 
-    Base.ctranspose(qarr::QuArray) = CTranspose(qarr)
-    Base.ctranspose(ct::CTranspose) = ct.qarr
+    #Base.ctranspose(qarr::QuArray) = CTranspose(qarr)
+    function ctranspose(qarr::QuArray)
+        ctcoeffs = ctranspose(rawcoeffs(qarr))
+        return CTranspose(QuArray(ctcoeffs,bases(qarr)))
+    end
+
+    #Base.ctranspose(ct::CTranspose) = ct.qarr
 
 ######################
 # Printing Functions #


acroy:
Good thinking, but actually this is on purpose. The `CTranspose` type is basically just a label, which denotes a `QuArray` as "ctransposed". The actual transposition is only done when necessary, e.g. in `getindex` or in `*` (note the use of `Ac_mul_B` in `*(::CTranspose, ::QuArray)`).

The printing of the coefficients is broken though...

jrevels:
This is intentional behavior. Our ctranspose is *lazy*, meaning that the actual conjugation and transposition on the coefficient array does not occur until the elements stored within are accessed (see how `getindex` and `setindex!` perform the eager ctranspose for access/mutation purposes, but do not perform the eager ctranspose on the underlying coefficient array). 

For a majority of the operations we care about, it is enough to simply mark a QuArray as conjugate transposed rather than to actually perform the conjugate transpose, which requires a full copy of the array. A good example of this is all of the recently defined multiplication code, which uses `Ac_mul_B` and the like.

What you got caught up on here, I think, is that the current pretty-printing methods are just shoddy - the only indication that `c_qm` has been conjugate transposed is that it prints out `DualMatrix` instead of `QuMatrix`. The coefficients printed, however, are simply the original ones obtained by using `rawcoeffs`. This is confusing and should be fixed in the future, but it's been lower priority so far compared to getting most other basic functionality up and running.

acroy:
@jrevels : only by 4 minutes :-)

amitjamadagni:
@acroy @jrevels thanks for the explanation on this. But I have a doubt given an operator/vector how do we end up with the `ctranspose` of it in sense ket -> bra and bra -> ket and operator to a operator. I would like to do something like `ctranspose(A)` where `A` is an operator and end up with the result.  Did I miss something really basic here ??

acroy:
Well, it depends of course want you want to do, but for all practical purposes `qt=qa'` for `qa::QuArray` behaves like the dual of `qa` (at least that is the plan). For example if you access the coefficients via `qt[indx]` you will find that they are conjugated (and transposed). Also if you write `qt*qa` you will get a number if `qa` is a QuVector and so forth. What are you trying to do?

However, we should probably add
```
coeffs(qa::QuArray)=qa.coeffs
coeffs(qt::CTranspose)=rawcoeffs(qt)'
```
to make life a bit easier? (But we should avoid to use `coeffs` too much, since it spoils the lazy evaluation)

amitjamadagni:
Yeah I was looking for the behavior of coeffs which is has been implemented in the latest pull.

PR #22, opened 2015-03-29
WIP : Spin Operators, Tests. WD : +,-,ops added.

amitjamadagni:
WIP : 
- [x] Spin operators to be implemented
- [x] Tests

Implemented 
- [x] +,- on QuArray, CTranspose
- [x] momentm_op, position_op added.

### DIFF
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -61,7 +61,7 @@ end
 
 *(dm1::DualMatrix, dm2::DualMatrix) = (dm1.qarr*dm2.qarr)'
 
-function +{B<:OrthonormalBasis,N}(qarr1::Union(QuArray{B,N},CTranspose{B,N}), qarr2::Union(QuArray{B,N},CTranspose{B,N}))
+function +{B<:OrthonormalBasis,N}(qarr1::AbstractQuArray{B,N}, qarr2::AbstractQuArray{B,N})
     if bases(qarr1) == bases(qarr2)
         return QuArray(coeffs(qarr1)+coeffs(qarr2), bases(qarr1))
     else
@@ -69,7 +69,7 @@ function +{B<:OrthonormalBasis,N}(qarr1::Union(QuArray{B,N},CTranspose{B,N}), qa
   end
 end
 
-function -{B<:OrthonormalBasis,N}(qarr1::Union(QuArray{B,N},CTranspose{B,N}), qarr2::Union(QuArray{B,N},CTranspose{B,N}))
+function -{B<:OrthonormalBasis,N}(qarr1::AbstractQuArray{B,N}, qarr2::AbstractQuArray{B,N})
     if bases(qarr1) == bases(qarr2)
         return QuArray(coeffs(qarr1)-coeffs(qarr2), bases(qarr1))
     else
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -61,8 +61,21 @@ end
 
 *(dm1::DualMatrix, dm2::DualMatrix) = (dm1.qarr*dm2.qarr)'
 
-+(qarr1::Union(QuArray,CTranspose), qarr2::Union(QuArray,CTranspose)) = bases(qarr1) == bases(qarr2) ? (return QuArray(coeffs(qarr1)+coeffs(qarr2), bases(qarr1))) : "Not compatible"
--(qarr1::Union(QuArray,CTranspose), qarr2::Union(QuArray,CTranspose)) = bases(qarr1) == bases(qarr2) ? (return QuArray(coeffs(qarr1)-coeffs(qarr2), bases(qarr1))) : "Not compatible"
+function +{B<:OrthonormalBasis,N}(qarr1::Union(QuArray{B,N},CTranspose{B,N}), qarr2::Union(QuArray{B,N},CTranspose{B,N}))
+    if bases(qarr1) == bases(qarr2)
+        return QuArray(coeffs(qarr1)+coeffs(qarr2), bases(qarr1))
+    else
+        error("Bases not compatible")
+  end
+end
+
+function -{B<:OrthonormalBasis,N}(qarr1::Union(QuArray{B,N},CTranspose{B,N}), qarr2::Union(QuArray{B,N},CTranspose{B,N}))
+    if bases(qarr1) == bases(qarr2)
+        return QuArray(coeffs(qarr1)-coeffs(qarr2), bases(qarr1))
+    else
+        error("Bases not compatible")
+  end
+end
 
 # scaling
 Base.scale!(num::Number, qarr::QuArray) = (scale!(num, rawcoeffs(qarr)); return qarr)
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -61,6 +61,9 @@ end
 
 *(dm1::DualMatrix, dm2::DualMatrix) = (dm1.qarr*dm2.qarr)'
 
++(qarr1::Union(QuArray,CTranspose), qarr2::Union(QuArray,CTranspose)) = bases(qarr1) == bases(qarr2) ? (return QuArray(coeffs(qarr1)+coeffs(qarr2), bases(qarr1))) : "Not compatible"
+-(qarr1::Union(QuArray,CTranspose), qarr2::Union(QuArray,CTranspose)) = bases(qarr1) == bases(qarr2) ? (return QuArray(coeffs(qarr1)-coeffs(qarr2), bases(qarr1))) : "Not compatible"
+
 # scaling
 Base.scale!(num::Number, qarr::QuArray) = (scale!(num, rawcoeffs(qarr)); return qarr)
 Base.scale!(num::Number, ct::CTranspose) = CTranspose(scale!(num', ct.qarr))
@@ -90,18 +93,18 @@ normalize(qarr::Union(QuArray,CTranspose)) = normalize!(copy(qarr))
 
 # General tensor product definitions for orthonormal bases
 function tensor{B<:OrthonormalBasis,T1,T2,N}(qarr1::AbstractQuArray{B,T1,N}, qarr2::AbstractQuArray{B,T2,N})
-    return QuArray(kron(coeffs(qarr1), coeffs(qarr2)), 
+    return QuArray(kron(coeffs(qarr1), coeffs(qarr2)),
                    map(tensor, bases(qarr1), bases(qarr2)))
 end
 
 function tensor{B<:OrthonormalBasis,T1,T2,N}(qarr1::CTranspose{B,T1,N}, qarr2::CTranspose{B,T2,N})
-    return QuArray(kron(rawcoeffs(qarr1), rawcoeffs(qarr2)), 
+    return QuArray(kron(rawcoeffs(qarr1), rawcoeffs(qarr2)),
                    map(tensor, rawbases(qarr1), rawbases(qarr2)))'
 end
 
 function tensor{B<:OrthonormalBasis}(ket::QuVector{B}, bra::DualVector{B})
-    return QuArray(kron(coeffs(ket), coeffs(bra)), 
-                   bases(ket,1), 
+    return QuArray(kron(coeffs(ket), coeffs(bra)),
+                   bases(ket,1),
                    bases(bra,1))
 end
 
--- a/src/arrays/ladderops.jl
+++ b/src/arrays/ladderops.jl
@@ -29,12 +29,12 @@
 
 function positionop(n::Int)
     cop = raiseop(n)
-    return (cop+cop')/sqrt(2.)
+    return scale!(1/sqrt(2.),cop+cop')
 end
 
 function momentumop(n::Int)
     cop = raiseop(n)
-    return im*(cop-cop')/sqrt(2.)
+    return scale!(1/sqrt(2.), im*(cop-cop'))
 end
 
 export raiseop,
--- a/src/arrays/ladderops.jl
+++ b/src/arrays/ladderops.jl
@@ -38,16 +38,24 @@ function momentumop(n::Int)
 end
 
 function spin_h1(j::Float64)
-    m = reverse([-j:j-1])
-    N = length(m)+1
-    m_coeffs = [sqrt(j*(j+1.0) - (x+1.0)*x) for x in m]
-    return spdiagm(m_coeffs,1,N,N)
+    if (isinteger(2*j) && j>0)
+        m = reverse([-j:j-1])
+        N = length(m)+1
+        m_coeffs = [sqrt(j*(j+1.0) - (x+1.0)*x) for x in m]
+        return spdiagm(m_coeffs,1,N,N)
+    else
+        return error("j must be non-negative integer or half integer")
+  end
 end
 
 function spin_h2(j::Float64)
-    m = reverse([-j:j])
-    N = length(m)
-    return spdiagm(m,0,N,N)
+    if (isinteger(2*j) && j>0)
+        m = reverse([-j:j])
+        N = length(m)
+        return spdiagm(m,0,N,N)
+    else
+        return error("j must be non-negative integer or half integer")
+  end
 end
 
 spin_jx(j::Float64) = QuArray(0.5*(spin_h1(j)+spin_h1(j)'))
@@ -71,4 +79,3 @@ export raiseop,
     sigmax,
     sigmay,
     sigmaz
-
--- a/src/arrays/ladderops.jl
+++ b/src/arrays/ladderops.jl
@@ -1,11 +1,11 @@
 ###########################
 # Ladder Operator Methods #
-########################### 
-    # n specifies which particle (a.k.a tensor product 
+###########################
+    # n specifies which particle (a.k.a tensor product
     # factor) the operator acts on
     raiseop(b::AbstractBasis, n=1) = QuArray(raisematrix(size(b), n), b, b)
     raiseop(lens, n=1) = raiseop(FiniteBasis(lens), n)
-    
+
     lowerop(b::AbstractBasis, n=1) = QuArray(lowermatrix(size(b), n), b, b)
     lowerop(lens, n=1) = lowerop(FiniteBasis(lens), n)
 
@@ -22,10 +22,22 @@
     before_eye(lens, pivot) = speye(prod(lens[[1:pivot-1]]))
     after_eye(lens, pivot) = speye(prod(lens[[pivot+1:length(lens)]]))
     function eye_sandwich(lens, pivot, op)
-        return kron(before_eye(lens, pivot), 
-                    op, 
+        return kron(before_eye(lens, pivot),
+                    op,
                     after_eye(lens, pivot))
     end
 
+function positionop(n::Int)
+    cop = raiseop(n)
+    return (cop+cop')/sqrt(2.)
+end
+
+function momentumop(n::Int)
+    cop = raiseop(n)
+    return im*(cop-cop')/sqrt(2.)
+end
+
 export raiseop,
-    lowerop
+    lowerop,
+    positionop,
+    momentumop
--- a/src/arrays/ladderops.jl
+++ b/src/arrays/ladderops.jl
@@ -37,45 +37,7 @@ function momentumop(n::Int)
     return scale(im/sqrt(2.), cop-cop')
 end
 
-function spin_h1(j::Float64)
-    if (isinteger(2*j) && j>0)
-        m = reverse([-j:j-1])
-        N = length(m)+1
-        m_coeffs = [sqrt(j*(j+1.0) - (x+1.0)*x) for x in m]
-        return spdiagm(m_coeffs,1,N,N)
-    else
-        return error("j must be non-negative integer or half integer")
-  end
-end
-
-function spin_h2(j::Float64)
-    if (isinteger(2*j) && j>0)
-        m = reverse([-j:j])
-        N = length(m)
-        return spdiagm(m,0,N,N)
-    else
-        return error("j must be non-negative integer or half integer")
-  end
-end
-
-spin_jx(j::Float64) = QuArray(0.5*(spin_h1(j)+spin_h1(j)'))
-spin_jy(j::Float64) = QuArray(-0.5*im*(spin_h1(j)-spin_h1(j)'))
-spin_jz(j::Float64) = QuArray(spin_h2(j))
-spin_jp(j::Float64) = QuArray(spin_h1(j))
-spin_jm(j::Float64) = QuArray(spin_h1(j)')
-sigmax = 2.0 * spin_jx(0.5)
-sigmay = 2.0 * spin_jy(0.5)
-sigmaz = 2.0 * spin_jz(0.5)
-
 export raiseop,
     lowerop,
     positionop,
-    momentumop,
-    spin_jx,
-    spin_jy,
-    spin_jz,
-    spin_jp,
-    spin_jm,
-    sigmax,
-    sigmay,
-    sigmaz
+    momentumop
--- a/src/arrays/ladderops.jl
+++ b/src/arrays/ladderops.jl
@@ -37,7 +37,46 @@ function momentumop(n::Int)
     return scale(im/sqrt(2.), cop-cop')
 end
 
+function spin_h1(j)
+    m = reverse([-j:j-1])
+    N = length(m)+1
+    m_coeffs = [sqrt(j*(j+1.0) - (x+1.0)*x) for x in m]
+    return spdiagm(m_coeffs,1,N,N)
+end
+
+function spin_h2(j)
+    m = reverse([-j:j])
+    N = length(m)
+    return spdiagm(m,0,N,N)
+end
+
+function spin_jx(j)
+    return QuArray(0.5*(spin_h1(j)+spin_h1(j)'))
+end
+
+function spin_jy(j)
+    return QuArray(-0.5*im*(spin_h1(j)-spin_h1(j)'))
+end
+
+function spin_jz(j)
+    return QuArray(spin_h2(j))
+end
+
+function spin_jp(j)
+    return QuArray(spin_h1(j))
+end
+
+function spin_jm(j)
+    return QuArray(spin_h1(j)')
+end
+
 export raiseop,
     lowerop,
     positionop,
-    momentumop
+    momentumop,
+    spin_jx,
+    spin_jy,
+    spin_jz,
+    spin_jp,
+    spin_jm
+
--- a/src/arrays/ladderops.jl
+++ b/src/arrays/ladderops.jl
@@ -37,38 +37,27 @@ function momentumop(n::Int)
     return scale(im/sqrt(2.), cop-cop')
 end
 
-function spin_h1(j)
+function spin_h1(j::Float64)
     m = reverse([-j:j-1])
     N = length(m)+1
     m_coeffs = [sqrt(j*(j+1.0) - (x+1.0)*x) for x in m]
     return spdiagm(m_coeffs,1,N,N)
 end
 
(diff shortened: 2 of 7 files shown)


src/arrays/arraymath.jl
```diff
@@ -61,6 +61,9 @@ end
 
 *(dm1::DualMatrix, dm2::DualMatrix) = (dm1.qarr*dm2.qarr)'
 
++(qarr1::Union(QuArray,CTranspose), qarr2::Union(QuArray,CTranspose)) = bases(qarr1) == bases(qarr2) ? (return QuArray(coeffs(qarr1)+coeffs(qarr2), bases(qarr1))) : "Not compatible"
```
acroy: I think you want to use `error("Bases not compatible.")` here. Should we use type parameters (for the basis and `N`) to restrict the functions in the first place?

amitjamadagni: @acroy something like
```julia
function +{B<:AbstractBasis,N}(qarr1::Union(QuArray{B,N}, CTranspose{B,N}), qarr2::Union(QuArray{B,N},CTranspose{B,N}))
```

acroy: Yes, that way we get e `MethodError` "for free" if someone tries to add say a QuVector and a QuMatrix. Nevertheless we have to check that the bases are the same (I am not sure if `==` works though).

jrevels: I think `==` is the right thing to do here, since it can be overloaded to work properly for each basis type - "value" equality is what we're looking for here, yes? Also, all we need to do to support mixed-basis type operations is to simply define some promotion rules between the various basis types, of which there aren't that many.

src/arrays/ladderops.jl
```diff
@@ -22,10 +22,22 @@
...
-        return kron(before_eye(lens, pivot), 
-                    op, 
+        return kron(before_eye(lens, pivot),
+                    op,
                     after_eye(lens, pivot))
     end
 
+function positionop(n::Int)
+    cop = raiseop(n)
+    return (cop+cop')/sqrt(2.)
```
acroy: We should use `scale!` here since `+` already makes a copy and `/` would make another one (same for `momentumop` below).

src/arrays/ladderops.jl
```diff
@@ -22,10 +22,22 @@
...
+end
+
+function momentumop(n::Int)
+    cop = raiseop(n)
+    return im*(cop-cop')/sqrt(2.)
+end
+
```
jrevels: A couple of points here:

1. We should support define these operator functions so that they can target specific factors of product states. See how `raiseop` and `lowerop` for an example of this; you could just reuse `eye_sandwhich` to make it work. Eventually, though, we should replace the usage of that method with specialized implementations that don't make temporary arrays. 

2. I believe it would be more efficient to scale the resulting `QuArray`s in place than make a copy, e.g.
```julia
function momentumop(n::Int)
    cop = raiseop(n)
    return scale!(im/sqrt(2.), cop-cop')
end
```

src/arrays/arraymath.jl
```diff
@@ -61,6 +61,22 @@ end
...
+end
+
+function -{B<:OrthonormalBasis,N}(qarr1::Union(QuArray{B,N},CTranspose{B,N}), qarr2::Union(QuArray{B,N},CTranspose{B,N}))
+    if bases(qarr1) == bases(qarr2)
+        return QuArray(coeffs(qarr1)-coeffs(qarr2), bases(qarr1))
+    else
+        error("Bases not compatible")
+  end
+end
+
```
jrevels: I know that we already have this pattern in the code, but maybe we should be annotating arguments as `AbstractQuArray{B,N}` instead of `Union(QuArray{B,N},CTranspose{B,N})` for functions like these. It's both less wordy and more generic. 

Thoughts, @acroy?

src/arrays/ladderops.jl
jrevels: @acroy Ah! Once again, I comment only to find that you've beat me to it by the time I'm done writing. I'm starting to believe that you are, in fact, a ninja...

jrevels: Actually, I just realized we won't be able to perform an in-place scaling of the Float64 array by a complex number; see my edit on my comment below.

src/arrays/ladderops.jl
amitjamadagni: @jrevels if I my understanding of the first comment is right  it meant that we should be able to track intermediate steps in producing the final answer. In specific have functions which produce the intermediate results and reuse them as necessary ?? But in these case it is a straight forward implementation. Please do correct me if I am missing something.

src/arrays/arraymath.jl
amitjamadagni: I guess we could, as QuArray and CTranspose are subtypes of AbstractQuArray. But could we compare bases if we used such a construct ??

jrevels: Yes. Your new method signature for `-`, for example, would look like this:

```julia
-{B<:OrthonormalBasis,N}(qarr1::AbstractQuArray{B,N},qarr2::AbstractQuArray{B,N})
```
...and the `bases` function is defined for both `QuArray` and `CTranspose`.

acroy: @jrevels : That would be nicer. I guess we haven't really specified the interface for `AbstractQuArray`s yet, but this would mean all types `<:AbstractQuArray` should support `coeffs` and `rawcoeffs` etc, right? Then we can generalize all the functions as you suggest and it will be easy to add new types with some specialized behavior (density matrices!).

amitjamadagni:
@jrevels  I was thinking a bit off on it. I realize it is costlier. Anyways thanks for this :).

acroy:
For spin operators we might want to keep an eye on [FixedSizeArrays.jl](https://github.com/SimonDanisch/FixedSizeArrays.jl) (also JuliaLang/julia#7568)?

src/arrays/arraymath.jl
jrevels: Yep - I think `bases`/`rawbases` should be in there as well.

acroy: I agree. That should be a separate PR and maybe we should also document the interface somewhere.

amitjamadagni:
@acory I was doing something like this based on QuTiP :
```julia
function spin_j(j)
          m = reverse([-j:j-1])
          N = length(m)+1
          m_coeffs = [sqrt(j*(j+1.0) - (x+1.0)*x) for x in m]
          return spdiagm(m_coeffs,1,N,N)
end
```
As far as my understanding of FixedSizeArrays.jl goes we are trying to create the array in a more efficient way with focus on the size. I am still trying to further understand but please do let me know if this is going in the right direction.

acroy:
@amitjamadagni : for now we probably want (have) to use sparse matrices as containers as you do. I just stumbled across FixedSizeArrays and thought those could be useful for stuff like Pauli spin matrices, which are quite common. AFAICT the FixedSizeArrays need Julia 0.4 anyways, so this is something for QuBase 2.0 :-)

amitjamadagni:
@acroy a review would be helpful. I have added the spin operators. Thanks

acroy:
You have to rebase against the latest master and maybe we should put the spin operators into a separate file?

amitjamadagni:
@acroy sure okay. But could we move the entire machinery into a single file with the name as operators.jl ??

src/arrays/ladderops.jl
```diff
@@ -22,10 +22,60 @@
...
+    cop = raiseop(n)
+    return scale!(1/sqrt(2.),cop+cop')
+end
+
+function momentumop(n::Int)
+    cop = raiseop(n)
+    return scale(im/sqrt(2.), cop-cop')
+end
+
+function spin_h1(j::Float64)
```
acroy: Maybe it is more elegant to use `Rational`s here (see [docs](http://julia.readthedocs.org/en/latest/manual/complex-and-rational-numbers/#rational-numbers))?

src/arrays/ladderops.jl
```diff
@@ -22,10 +22,60 @@
...
+        return error("j must be non-negative integer or half integer")
+  end
+end
+
+spin_jx(j::Float64) = QuArray(0.5*(spin_h1(j)+spin_h1(j)'))
+spin_jy(j::Float64) = QuArray(-0.5*im*(spin_h1(j)-spin_h1(j)'))
+spin_jz(j::Float64) = QuArray(spin_h2(j))
+spin_jp(j::Float64) = QuArray(spin_h1(j))
+spin_jm(j::Float64) = QuArray(spin_h1(j)')
+sigmax = 2.0 * spin_jx(0.5)
```
acroy: Maybe `sigma_x` etc for consistency? (We could even use Unicode sigma in addition). And you should declare them const.

acroy:
Let's see what Jarrett says, he has a better overview of the repo structure. I thought we could move the spin operators into a file spinops.jl or such ...

amitjamadagni:
@acroy I am trying to rebase against the latest master. In the attempt I created a bit of mess with the commits. Any pointers relating to rebase would be helpful. Thanks.

jrevels:
I agree with using `Rational`s for the spin operators and moving them to their own file ("spinops.jl" sounds good to me). 

@amitjamadagni You might've already seen this before, but I found [this tutorial](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) very helpful when learning to rebase. Is there a specific thing you're having trouble with, error message, etc.?

amitjamadagni:
@jrevels Yeah I have been trying and playing around but the problem I end up is I am unable to see the previous commits when I do a rebase -i master, to squash. I will give it a try once again and let you know.

amitjamadagni:
@acroy and also it would helpful to hear your comments on the unicode method names. I have been able to generate sigma_x but not sigma_y and sigma_z (I guess there is no support for \sigma\_y and \sigma\_z).

acroy:
Regarding the Rationals I actually thought to use them in `spin_h1/2` instead of the isinteger test. It somehow feels strange to allow arbitrary floats for the spin. The matrices would of course (?) have floating point elements (convert `m` to float before constructing the sparse matrix). 

I haven't looked too much into the Unicode stuff yet. If the other sigmas are not working, maybe we should look at this at some other point (it's not crucial, but just nice to have).

amitjamadagni:
@acroy I guess we would still require the isinteger test as a check on the input, just having a rational type would not help as we can still have inputs like 4/3 which are not allowed and also we do not have a `isrational` method to check. Please do let me know if I am missing something here. Also this is a better check on the floats that we are using, I guess there are no edge cases which I can think of so. Please do let me know your thoughts on it. Thanks.

acroy:
I thought we could test the numerator for being positive and the denominator for being 1 or 2. Might be that isinteger for Rationals actually does something like that. I don't insist on using Rationals though, it just feels odd to allow arbitrary floats and but then restrict the input (what happens if you put in 0.500001?)

amitjamadagni:
@acroy I just ran some benchmarks. Here are the results :
Using Rational we get :
0.002287712 seconds (17960 bytes allocated)
Using Float we get :
0.002204122 seconds (17936 bytes allocated)

Any thoughts on this would be helpful.

acroy:
I wouldn't expect a big performance difference for that piece of code. If we would use Rationals as elements of the matrices, it would certainly have an impact, since (almost) all linear algebra routines are optimized for floats. 

One might get a (negligible?) performance improvement if one passes numerator and denominator as separate integers to those routines. If both values are knows at compile time, Julia might be able to optimize the functions (eliminate branches for instance). But I don't think that it is crucial, unless we expect those routines to be called millions of times (although then garbage collection will kill us first).

My point was only that it is somewhat unintuitive/confusing to be able to pass arbitrary floats as spins. Moreover, there might actually be edge case, like passing `0.500001` or such.

jrevels:
I can understand passing the numerator or denominator separately.

But if we do go for allowing a single number argument for spin, I think it would be odd for user to enter something like `spin_jx(1/2)` and have it throw an error.

I agree that we should test for edge cases like @acroy (and maybe start increasing/decreasing precision until we it hits a failure mode). I'd do it myself but I'm in transit right now...

amitjamadagni:
@acroy my bad I had forgot to mention that the input check was for Rational but then I had converted the internals to float. The memory used is more in case of rationals when compared to float, can this be a case for consideration ?? Also from my understanding we should be setting the precision so that we consider the edge cases, right ??

acroy:
We should check what `isinteger` is doing under the hood. Maybe it already uses `round` or such and we don't have to worry about the edge cases...

A more sophisticated (but probably overly complicated) approach might be something like
```
abstract Spin{S} # S is just the numerator of the spin value
immutable HalfSpin{S} <: Spin{S} end
immutable IntegerSpin{S} <: Spin{S} end
```
plus constructors and some convenience functions. Then we exactly know what kind of spin we have to deal with. Of course the problem would be shifted to implementing the constructor taking a float (or conversion from float), but this could be fixed/changed later without affecting anything else. The question is if we need those types for something else? Anyways it might be a nice exercise.

@amitjamadagni Where those timings obtained by running the function once? If so you basically measure compilation time/memory. Have a look at the [performance tips](http://julia.readthedocs.org/en/latest/manual/performance-tips/) to read an explanation. If you ran it twice already, just ignore me ;- )

amitjamadagni:
@acroy the instances of `isinteger` in `floatfuncs.jl` and `rational.jl` from base are as follows :

```julia
isinteger(x::FloatingPoint) = (trunc(x)==x)&isfinite(x)

isinteger(x::Rational) = x.den == 1
```
I have searched for `trunc` but could not find correct references to it. I will work on the other idea as well and send in a commit.

amitjamadagni:
@acroy I have made an attempt at implementing immutable types `HalfSpin` and `IntegerSpin`. Any thoughts on this would be helpful.

src/arrays/spinops.jl
```diff
@@ -0,0 +1,124 @@
...
+#const σₓ = sigma_x
+
+#########################
+# Implementation  -  2  #
+#########################
+
+abstract Spin{S}
+
+immutable HalfSpin{S} <: Spin{S}
+    num::S
```
acroy: You don't need to store `den`, since you can assume that `den==2` because that information is contained in the type name (HalfSpin!). And actually you don't have to store `num` either since you can use the type parameter `S` for that (remember type parameters can be integers).

src/arrays/spinops.jl
```diff
@@ -0,0 +1,124 @@
...
+    j = float(k.num/k.den)
+    return mat_coeffs_2(j)
+end
+
+function spin_h2(k::IntegerSpin)
+    j = float(k.num)
+    return mat_coeffs_2(j)
+end
+
+spin_jx(j::Union(HalfSpin,IntegerSpin)) = QuArray(0.5*(spin_h1(j)+spin_h1(j)'))
```
acroy: Instead of the `Union` you can just say `j::Spin`. Since `Spin` is abstract the function will apply to all concrete subtypes; in this case `HalfSpin` and `IntegerSpin`.

src/arrays/spinops.jl
```diff
@@ -0,0 +1,124 @@
...
+spin_jy(j::Union(HalfSpin, IntegerSpin)) = QuArray(-0.5*im*(spin_h1(j)-spin_h1(j)'))
+spin_jz(j::Union(HalfSpin, IntegerSpin)) = QuArray(spin_h2(j))
+spin_jp(j::Union(HalfSpin, IntegerSpin)) = QuArray(spin_h1(j))
+spin_jm(j::Union(HalfSpin, IntegerSpin)) = QuArray(spin_h1(j)')
+spin_half = HalfSpin(1,2)
+const sigma_x = 2.0 * spin_jx(spin_half)
+const sigma_y = 2.0 * spin_jy(spin_half)
+const sigma_z = 2.0 * spin_jz(spin_half)
+
+export HalfSpin,
```
acroy: Unless we are sure we want to use them somewhere else, it might be better to not export those types.

src/arrays/spinops.jl
```diff
@@ -0,0 +1,124 @@
...
+    IntegerSpin,
+    spin_jx,
+    spin_jy,
+    spin_jz,
+    spin_jp,
+    spin_jm,
+    sigma_x,
+    sigma_y,
+    sigma_z,
+    σₓ
```
acroy: This one doesn't exist anymore.

src/arrays/spinops.jl
amitjamadagni: Ah, we did the same in addition of QuArray in one of the previous commits !!! I will refine it.

acroy: Actually you probably don't have to specify the argument type at all for those functions. We should instead make sure that we have all the constructors and conversions in place, which will then get called automatically.

src/arrays/spinops.jl
amitjamadagni: I forgot to add it. Sorry for this.

acroy:
Cool! That looks basically like I anticipated it. We will need some constructors/conversions, say from float to HalfSpin, but for this we could just take your previous implementation.

amitjamadagni:
@acroy I will send in a commit with the implementation 1 and then I guess this would be ready for merge ??

acroy:
You only need your precious approach for the `HalfSpin` constructor and the we will see. Don't forget to add tests. Maybe using commutator relations might be a good approach for testing?

amitjamadagni:
@acroy sorry I could not get that comment :(

acroy:
Sorry, I meant the previous implementation (isinteger)  only applies to the HalfSpin constructor, which takes a float/Rational as input. If you have that and take the other comments into account we will see where we are. Moreover, you will need some tests. One way might be to use the commutation relations (simplest case are Pauli matrices) as tests.

amitjamadagni:
@acroy I have made an attempt at the following :
* Type independence of Spin
   * Type conversion to spintypes
* Commutator method
* Tests using the commutator method

I have tried to work around on the making the HalfSpin type without den but I am unable to crack the problem. Any leads on this would be helpful.

src/arrays/arraymath.jl
```diff
@@ -90,23 +106,32 @@ normalize(qarr::Union(QuArray,CTranspose)) = normalize!(copy(qarr))
...
                    bases(bra,1))
 end
 
 tensor{B<:OrthonormalBasis}(bra::DualVector{B}, ket::QuVector{B}) = tensor(ket, bra)
 
+##############
+# Commutator #
+##############
+
+function commutator{B<:OrthonormalBasis}(qm1::Union(QuMatrix{B},DualMatrix{B}), qm2::Union(QuMatrix{B},DualMatrix{B}))
```
acroy: I think `AbstractQuMatrix` instead of the `Union` would be better.

test/operatortest.jl
```diff
@@ -0,0 +1,8 @@
+#######################
+# Spin Operators Test #
+#######################
+
+@assert commutator(sigma_x, sigma_y) == 2*im*sigma_z
```
acroy: This is really nice! Could you add a similar test for position and momentum operators (just use some reasonable basis size).

acroy:
Regarding the `Spin` types I thought something like this might work
```julia
abstract Spin{S}

immutable HalfSpin{S} <: Spin{S} end
immutable IntegerSpin{S} <: Spin{S} end

# constructors
HalfSpin(j::Integer) = j > 0 ? HalfSpin{j}() : error("Spin must be positive")

function HalfSpin(j::Float64)
    if (j > 0) && !isinteger(j) && isinteger(j*2)
        HalfSpin(int(j*2))
    else
        error("Spin must be positive and a multiple of 1/2.")
    end
end

IntegerSpin(j::Integer) = j > 0 ? IntegerSpin{j}() : error("Spin must be positive")

function IntegerSpin(j::Float64)
    if (j > 0) && isinteger(j)
        IntegerSpin(int(j))
    else
        error("Spin must be positive and integer.")
    end
end

# conversion
Base.convert{S}(::Type{IntegerSpin{S}}, j::Integer) = IntegerSpin(j)
Base.convert{S}(::Type{HalfSpin{S}}, j::Float64) = HalfSpin(j)

# extracting the spin
spin{S}(is::IntegerSpin{S}) = S
spin{S}(his::HalfSpin{S}) = S/2
```

acroy:
@jrevels : Any thoughts on the `Spin` types. Are they overkill? We could also go back to Amit's initial approach and introduce the spins separately...

amitjamadagni:
@acroy I have tried to play around with the code, here are few doubts that I have :

1. Why is it that we need convert, if we have already defined the methods for integer and float,  we have something like 
```julia
hs = HalfSpin(1/2)
#giving out 
HalfSpin{1}()
``` 
which is what is expected out of convert.

2. I have tried this 

```julia 
x = IntegerSpin(2.0)
println(x)
println(Spin(x))
```
This gives an error like 
```ijulia
ERROR: cannot define function Spin; it already has a value
```
I guess I am missing something here. Please do let me know what I am missing. Thanks.

acroy:
@amitjamadagni : I think it is just a type, it should be `spin(x)`.

amitjamadagni:
@acroy in sense it fails to compile the file, giving out the error. Is it like we are extracting the value from the type ?? And also any thoughts on convert ??

acroy:
@amitjamadagni : Not sure what you mean, the code I posted works for me as expected:
```
julia> x = HalfSpin(1/2); println(x); println(spin(x))
HalfSpin{1}()
0.5

julia> x = IntegerSpin(1); println(x); println(spin(x))
IntegerSpin{1}()
1

julia> x = IntegerSpin(2.); println(x); println(spin(x))
IntegerSpin{2}()
2
```

acroy:
Maybe `convert` is not necessary. The problem is that we want to avoid functions, whose return **type** depends on the **value(s)** of the argument(s). This would cause type instability. I guess we should then rename the `spin` functions to something like `spin_value` and instead introduce
```
spin{S}(is::Spin{S}) = is 
spin(j::Integer) = IntegerSpin(j)
spin(j::Float64) = HalfSpin(j)
```
Those would be like your `convert_to_spintypes`. Then `spin(1.0)` wouldn't work of course, but OTOH there is no reason why a float should be accepted for this.

Now that I am writing this, I realize that the `IntegerSpin` type is actually superfluous. We can do everything in terms of `HalfSpin`:
```julia
abstract Spin{S}

immutable HalfSpin{S} <: Spin{S} end

# constructors
HalfSpin(j::Integer) = j > 0 ? HalfSpin{2*j}() : error("Spin must be positive")

function HalfSpin(j::Float64)
    if (j > 0) && isinteger(j)
        HalfSpin(int(j))
    elseif (j > 0) && isinteger(j*2)
        HalfSpin{int(2*j)}()
    else
        error("Spin must be positive and a multiple of 1/2.")
    end
end

# extracting the spin
spin_value{S}(his::HalfSpin{S}) = S/2

spin{S}(is::Spin{S}) = is 
spin(j::Integer) = HalfSpin(j)
spin(j::Float64) = HalfSpin(j)
```
(Maybe we don't even need the abstract Spin type). Now the type of all functions is independent of the actual value and everything converts the same way.

amitjamadagni:
@acroy I have added the latest implementation and modified other convenience functions. Yeah the doubts on convert method stand clarified. Also regarding the `error` I made a very grave mistake. Sorry for that.

acroy:
Don't worry, nobody is perfect :-) Let's see what Jarrett thinks of this, but I think this looks pretty much Ok (in the end we might want to remove Implementation 1 and squash the commits).

amitjamadagni:
@jrevels a review would be helpful. Thanks :)

src/arrays/spinops.jl
```diff
@@ -0,0 +1,110 @@
...
+    elseif (j > 0) && isinteger(j*2)
+        HalfSpin{int(2*j)}()
+    else
+        error("Spin must be positive and a multiple of 1/2.")
+    end
+end
+
+# extracting the spin
+spin_value{S}(his::HalfSpin{S}) = S/2
+
```
jrevels: You don't need to name the argument for this function, since it only depends on the argument's type:

```
spin_value{S}(::HalfSpin{S}) = S/2
```

I find that more stylistically pleasing, just because it's easier to see that the function isn't value-dependent.

src/arrays/spinops.jl
```diff
@@ -0,0 +1,110 @@
...
+    m_coeffs = [sqrt(j*(j+1.0) - (x+1.0)*x) for x in m]
+    return spdiagm(m_coeffs,1,N,N)
+end
+
+function mat_coeffs_2(j::Float64)
+    m = reverse([-j:j])
+    N = length(m)
+    return spdiagm(m,0,N,N)
+end
+
```
jrevels: `reverse([-j:j-1])` above can be replaced by `[j-1:-1:-j]`, which should be more memory efficient since you'll only need to pass through the range once (`reverse` also works on ranges, btw: `reverse(1:10)` --> `10:-1:1`).

src/arrays/spinops.jl
```diff
@@ -0,0 +1,110 @@
...
+function HalfSpin(j::Float64)
+    if (j > 0) && isinteger(j)
+        HalfSpin(int(j))
+    elseif (j > 0) && isinteger(j*2)
+        HalfSpin{int(2*j)}()
+    else
+        error("Spin must be positive and a multiple of 1/2.")
+    end
+end
+
```
jrevels: In the above, `j` moves from the value domain to the type domain, making this a type-unstable constructor. In addition, we multiply `j` by 2, which I find confusing for this constructor, since `HalfSpin{S}()` represents a spin value of `S * 1/2`. Thus, if I'm passing in `3` to a constructor for `HalfSpin`, I would expect that to return a representation of the spin value `3 *1/2`, rather than a total spin of `3`. I think this kind of doubling would be better to implement in the `spin` function below instead.

Also, the checks being done here become pointless, since I would bypass it just by calling `HalfSpin{2}()` instead of calling `HalfSpin(2)` (the former is also more natural to me anyway, since it's type-stable).

I say we should just remove the above constructors for `HalfSpin`, and just re-define the type like this:

```
immutable HalfSpin{S} 
    function HalfSpin()
        S > 0 ? new() : error("Spin must be positive")
    end
end
```

...which provides a single, safe, type-stable constructor.

Then, we can let the `spin` function below handle converting a numeric argument to the appropriate spin value (with additional checks, doubling, etc.). The `spin` function would still be type-unstable, since eventually the argument would pass from the value domain to the type domain, but I don't think it will matter that much as long as that function isn't directly used for heavy computations.

acroy: Yep, I fully agree.

@amitjamadagni : Please also remove the "old" implementation 1.

src/arrays/spinops.jl
acroy: +1

amitjamadagni:
@acroy  @jrevels I have made an attempt. A review would be helpful.

src/arrays/arraymath.jl
acroy: Actually we don't want to always return a `QuArray` here if we allow the arguments to be any subtype of `AbstractQuArray`. We probably need some promotion rules to get this right. But as we said above all of this should be done in a separate PR. For now it is Ok to leave it like that. @jrevels ?

acroy:
I think it is basically Ok. If you still have energy, you could add some comments to that `spin` function, in particular reminding us that it is not type stable. It would also be great if you could squash your commits.

amitjamadagni:
@acroy I am unable to squash some of the previous commits.

acroy:
Why?

amitjamadagni:
I dont get the previous 8 commits before the last when i do a git rebase -i master.

acroy:
It should probably be `git rebase -i HEAD~9`.

amitjamadagni:
@acroy Thanks for the help. Done squashing.

acroy:
Mhm, I can't merge it. Maybe you have to rebase again?

amitjamadagni:
@acroy seems like I am stuck in a loop. Can I close this and open an another pull request ??

PR #24, opened 2015-04-11
Operators, Tests.

amitjamadagni:
Added : 
Spin Operators, 
Related tests using Commutator,
Momentum and position operators.

Reference #22.


amitjamadagni:
@acroy I guess this is fine.

PR #25, opened 2015-04-17
Operators : Squeeze, displaceop. Addition : expm, full

amitjamadagni:
An attempt at implementing expm, full (sparse to dense) as expm is not implemented for sparse matrix. With respect to the implementation of expm with respect to sparse matrix, I have been referring to the https://github.com/JuliaLang/LinearAlgebra.jl/issues/174 which had hints at scipy implementation. The scipy implementation is based on the following algorithm (http://core.ac.uk/download/pdf/17664.pdf) as cited in the documentation. I will try a hand at implementing this if already not implemented, but that said as exponentiation is only available for dense matrix, I have implemented full function and added the expm on the top of that.  
Also the `*` method using DualMatrix was giving some inconsistent results, an attempt has been made to get it work.  
Any comments would be helpful. Thanks


src/arrays/arraymath.jl
```diff
@@ -90,6 +90,12 @@ Base.scale(qarr::Union(QuArray,CTranspose), num::Number) = scale(num, qarr)
 *(qarr::Union(QuArray,CTranspose), num::Number) = scale(qarr, num)
 /(qarr::Union(QuArray,CTranspose), num::Number) = scale(1/num, qarr)
 
+# sparse to dense
+Base.full(qarr::AbstractQuMatrix) = QuArray(full(coeffs(qarr)),bases(qarr))
+
+# exponential of dense matrix
+Base.expm(qarr::AbstractQuMatrix) = QuArray(expm(coeffs(full(qarr))),bases(qarr))
```
acroy: Wouldn't it be easier to just say `expm(full(coeffs(qarr)))`? But maybe it doesn't matter.

amitjamadagni: Yeah !!! Even I was thinking after I pushed. Let me check on the efficiency ...

src/arrays/ladderops.jl
```diff
@@ -37,7 +37,16 @@ function momentumop(n::Int)
     return scale(im/sqrt(2.), cop-cop')
 end
 
+squeeze_construct(a::AbstractQuMatrix,b::AbstractQuMatrix, z::Union(Complex64, Float64)) = 0.5*(z'*(a*b)-z*(a'*b'))
+squeezing(a::AbstractQuMatrix, b::AbstractQuMatrix, z::Union(Complex64, Float64)) = expm(squeeze_construct(a,b,z))
+squeeze(n::Int,z::Union(Complex64, Float64)) = squeezing(lowerop(n),lowerop(n),z)
+
+displaceop(n::Int,alpha::Union(Complex64, Float64)) = expm(alpha*lowerop(n)'-alpha'*lowerop(n))
```
acroy: I think it is better to use `raiseop(n)` instead of `lowerop(n)!`, which makes an additional copy. In fact it would be even better to define
```
function displacementop( nb::Int, alpha::Number )
	ad = scale!(alpha, raiseop( nb ))
	# displacement to location alpha
	return expm( ad - ad')
end
```

src/arrays/ladderops.jl
```diff
@@ -37,7 +37,16 @@ function momentumop(n::Int)
     return scale(im/sqrt(2.), cop-cop')
 end
 
+squeeze_construct(a::AbstractQuMatrix,b::AbstractQuMatrix, z::Union(Complex64, Float64)) = 0.5*(z'*(a*b)-z*(a'*b'))
```
acroy: I am not sure I understand this construct, but in QuDOS I have used the following
```
function squeezingop( nb::Int, z::Union(Complex128, Float64) )
	ad2 = 0.5*z*creationop( nb )*creationop( nb )
	sparse( expm( dense(ad2' - ad2) ) )
end
```
Do you think we will need the `squeezing` operation?

src/arrays/arraymath.jl
```diff
@@ -59,7 +59,7 @@ function *{B<:OrthonormalBasis}(qm1::QuMatrix{B}, qm2::QuMatrix{B})
                    bases(qm2,2))
 end
 
-*(dm1::DualMatrix, dm2::DualMatrix) = (dm1.qarr*dm2.qarr)'
+*(dm1::DualMatrix, dm2::DualMatrix) = QuArray(coeffs(dm1)*coeffs(dm2),bases(dm1,1), bases(dm2,2))
```
acroy: I guess this change is the reason why Travis fails. The old version returns a DualMatrix, while this returns a QuArray. I agree, that we should try to avoid accessing fields directly, but the old version is more concise. Maybe we need some function to access the array in a DualMatrix?

acroy:
I think for sparse matrices one mostly wants to avoid calculating `expm` directly, since the latter is typically dense and large. This is the reason why `empv` *et al* are so important. For those one rather computes `expm*v` where `v` is some vector. That said, I think that `expm(full(S))` is fine for small problems. If one is really serious about it, one needs something like `expmv` for `QuArrays` and functions that displace/squeeze/etc a vector. If we want to include such functions in QuBase, we can pull in Expokit.jl or ExpmV.jl for instance.

src/arrays/ladderops.jl
acroy: I have to take that back: `lowerop(n)!` doesn't make a copy, but for a normal array it would. However, multiplication with `alpha` creates copies.

src/arrays/arraymath.jl
jrevels: The removed version is also more efficient, AFAIK, since it doesn't have to take the conjugate transpose. 

But I suppose you're right about having an accessor function for the `qarr` field. The problem is `qarr` works really well as a descriptive, concise variable name, and I'd like to keep it if possible, so I wonder what to call the accessor function. Also, technically, `ctranspose` is basically implemented as an accessor function. But I suppose we should have an ACTUAL one, with which we can redefine `ctranspose` and all the other instances of direct field access. 

@amitjamadagni You said you were getting inconsistent behavior with the old definition; do you have an example?

jrevels:
@acroy At some point, we should probably include a definition like the following to dispatch on for sparse QuArrays:

```
typealias SparseQuArray{B,T,N,S<:AbstractSparseArray} QuArray{B,T,N,S}
```

> If we want to include such functions in QuBase, we can pull in Expokit.jl or ExpmV.jl for instance.

+1, and I think it makes sense to include those functions (in the form of `using` the packages, of course).

src/arrays/arraymath.jl
amitjamadagni: I had done this 
```ijullia
julia> lowerop(2)'*raiseop(2)'
2x2 DualMatrix in FiniteBasis{Orthonormal,1}:
...coefficients: SparseMatrixCSC{Float64,Int64}

	[1, 1]  =  1.0
```
I guess we should be getting 
```ijulia
julia> lowerop(2)'*raiseop(2)'
2x2 DualMatrix in FiniteBasis{Orthonormal,1}:
...coefficients: SparseMatrixCSC{Float64,Int64}

	[2, 2]  =  1.0
```

acroy: Good catch. It should be `*(dm1::DualMatrix, dm2::DualMatrix) = (dm2.qarr*dm1.qarr)'` since A'B'=(BA)'

acroy:
I would suggest to wait with using Expokit.jl et al. We have to look into it for the propagators anyways.

+1 to `SparseQuArray`.

src/arrays/arraymath.jl
jrevels: Ah! I didn't even notice that I had switched them, really nice catch. We should have a test for that...

amitjamadagni:
One build passes and the other build fails. The error says "does not pass tests in QuBase" for the second build. Any insights would be helpful.

src/arrays/arraymath.jl
amitjamadagni: Seems like `expm(full(ceoffs(qarr)))` is more efficient.

src/arrays/ladderops.jl
amitjamadagni: There is a mention of [generalized squeezing operator] (http://qutip.org/docs/3.0.1/apidoc/functions.html#qutip.operators.squeezing) on the QuTip docs. So I thought we could have this, as well we use it in the next construct which is a particular case.

acroy: Makes sense, but we can still call it `squeezingop` - multiple dispatch is our friend. Please change it to use `scale!` in order to avoid copies.

amitjamadagni: Sure. I have used scale in the latest commit.

acroy:
I guess the Travis failure is not related to this PR. Might be a consequence of JuliaLang/julia#10380. @jrevels : Any ideas?

amitjamadagni:
@acroy done with the edits.

acroy:
Looks Ok, but I want to wait for @jrevels to comment on the Travis failure. Could you add a test for `displaceop`? If one applies the displacement op to the ground state, one gets a [coherent state](http://en.m.wikipedia.org/wiki/Coherent_states) and you can test if the elements in the state vector are correct.

amitjamadagni:
@acroy I get an 
```julia
ERROR: InexactError()
 in setindex! at sparse/sparsematrix.jl:1253
 in generic_scale! at linalg/generic.jl:18
 ```
 when I use the combination of `complex` type for `alpha` parameter in `displaceop`. So I have replaced it with scale. I will work on this and see if this can be made to work if possible with scale.

**Comment by acroy** [comment]
I guess you have to use `@test_approx_eq` or similar. Why do you "hard code" the expected result instead of calculating `exp(-1/2)` or whatever for the coefficients?

**Comment by amitjamadagni** [comment]
@acroy we could have done something like 
```ijulia
exp(-0.5)*(statevec(1,FiniteBasis(2))+statevec(2,FiniteBasis(2)))
2-element QuVector in FiniteBasis{Orthonormal,1}:
...coefficients: Array{Float64,1}
[0.6065306597126334,0.6065306597126334]
```
which is not equal to the required output. (Truncated version)

We could define coherent state as 
```julia
coherent(n::Int,alpha::Number) = displaceop(n,alpha)*statevec(1,FiniteBasis(n))

# then test as
 
coherent(2,1) == displaceop(2,1)*statevec(1,FiniteBasis(2))
```

We could also have the expansion of coherent states i.e., `|alpha> = exp(-mod(alpha)^2)/2)*(sigma[(alpha)^n/sqrt(n!)|n>) `

acroy:
A function `coherentstatevec` should probably either be implemented using the `displace` function (which will use `expmv`) or directly using the expansion in Fock states. But your version is probably Ok for now.

Anyways, the test probably fails because your basis is too small. Using `QuDOS.coherentstatevec` I get
```
julia> cs = QuDOS.coherentstatevec(10,1.)
QuStateVec{Array{Complex{Float64},1}}(Complex{Float64}[0.606531+0.0im,0.606531+0.0im,0.428882+0.0im,0.247615+0.0im,0.123808+0.0im,0.0553686+0.0im,0.022603+0.0im,0.00854887+0.0im,0.00299672+0.0im,0.00110007+0.0im],10,[10])

julia> cs[1]
0.6065306597114603 + 0.0im

julia> exp(-0.5)
0.6065306597126334

julia> Base.Test.@test_approx_eq_eps real(cs[1]) exp(-0.5) 1e-6
```
You might be able to use a similar test if you increase the basis size.

amitjamadagni:
@acroy I have made an attempt at implementing `coherentstatevec` using expm as well the expanding the summation. I have added the tests for `displaceop` using the same. A review would be helpful.

src/arrays/constructors.jl
```diff
@@ -16,4 +16,16 @@
...
+function coherentstatevec(n::Int,alpha::Number;method="direct")
+    s = rand(n)
+    z = zeros(QuArray(s))
+    for i in [1:n]
+        z = z + ((alpha^(i-1))/sqrt(factorial(i)))*statevec(i, FiniteBasis(n))
+    end
+    return scale!(exp(-abs2(alpha)/2),z)
+end
+
+coherentstatevec(n::Int, alpha::Number) = displaceop(n,alpha)*statevec(1,FiniteBasis(n))
```
acroy: I don't think we need both versions and for now I would prefer the second one. There might actually be a problem having both, because they only differ in the keywords and I am not sure how Julia handles this. You can check by calling 
```
@which coherentstatevec(20,1,method = "direct")
@which coherentstatevec(20,1)
```
To see which versions are actually called.

test/operatortest.jl
```diff
@@ -10,11 +10,25 @@
...
 
 p = positionop(2)
 m = momentumop(2)
 @assert coeffs(commutator(sigma_x, p)) == spzeros(2,2)
 @assert coeffs(commutator(sigma_y, m)) == spzeros(2,2)
+
+csvd = coherentstatevec(20,1,method = "direct")
+csvo = displaceop(20,1)*statevec(1,FiniteBasis(20))
+@test_approx_eq_eps coeffs(csvd)[1] coeffs(csvo)[1] 1e-8
+@assert coherentstatevec(10,1.) == displaceop(10,1.)*statevec(1,FiniteBasis(10))
```
acroy: This test is redundant since the right hand side is the definition of your function...

src/arrays/constructors.jl
```diff
@@ -16,4 +16,16 @@
     one_at_ind!(arr, i) = setindex!(arr, one(eltype(arr)), i)
     single_coeff(i, lens...) = one_at_ind!(zeros(lens), i)
 
-export statevec
\ No newline at end of file
+function coherentstatevec(n::Int,alpha::Number;method="direct")
+    s = rand(n)
```
acroy: Just as an aside, why are you not using `zeros(n)` here? Calling `rand` will actually mean unnecessary computations. For small `n` this is fine, but for larger `n` you will notice the difference.

src/arrays/constructors.jl
amitjamadagni: I have tried this out 

```julia
println(@which coherentstatevec(10,1,method="direct"))
println(coherentstatevec(10,1,method="direct"))
println(@which coherentstatevec(10,1))
println(coherentstatevec(10,1))

# gives out 

coherentstatevec(n::Int64,alpha::Number) at QuBase.jl/src/arrays/constructors.jl:28
10-element QuVector in FiniteBasis{Orthonormal,1}:
...coefficients: Array{Float64,1}
[0.6065306597126334,0.6065306597126334,0.4288819424803534,0.24761510494160166,0.12380755247080083,0.05536842069051653,0.02260406309258736,0.008543532794657104,0.0030205949871958465,0.001006864995731949]
coherentstatevec(n::Int64,alpha::Number) at QuBase.jl/src/arrays/constructors.jl:28
10-element QuVector in FiniteBasis{Orthonormal,1}:
...coefficients: Array{Float64,1}
[0.6065306597114603,0.6065306597355303,0.4288819421804789,0.24761510797604577,0.12380752738214919,0.05536859516814174,0.02260302507996636,0.008548870306885925,0.0029967233467164565,0.0011000726308558629]

```
The output is different though they point to the same method which is reflected by the `@which`

src/arrays/constructors.jl
```diff
@@ -16,4 +16,16 @@
...
+    z = zeros(QuArray(s))
+    for i in [1:n]
+        z = z + ((alpha^(i-1))/sqrt(factorial(i)))*statevec(i, FiniteBasis(n))
```
acroy: ... And it should be `factorial(i-1)`. Note that creating the coefficients like that is probably not a good idea if n is large. If one would do this it might be better (I.e. More efficient) to use a recursive prescription. Something like (untestet!)
```
cs = zeros(n)
cs[1]=one(alpha)
For i in 2:n
   cs[i] = alpha/sqrt(i-1)*cs[i-1]
end
QuArray(cs)
```
Also note that the resulting state is not normalized (only approximately), because the formula for the elements in Fock space assumes infinite n. But I would anyways go with the other definition below and only use the formula in the test.

src/arrays/constructors.jl
acroy: To be sure you can put a `println` statement in one of the functions ...

src/arrays/constructors.jl
amitjamadagni: Yeah I just noticed the factorial(i-1), my bad. Thanks for this ... will implement using the recursive technique.

src/arrays/constructors.jl
amitjamadagni: I gave in a `println` for the `rand(n)` and this is the output ... seems like the method is being called but is not reflected by `@which`
```julia
coherentstatevec(n::Int64,alpha::Number) at QuBase.jl/src/arrays/constructors.jl:28

# printing rand(n)

[0.019244684430714143,0.3719091096873184,0.9047165798455694,0.6866397480495858,0.5561245580875909,0.14391135614954487,0.6180840544733457,0.002417775152962154,0.9427717780056364,0.20641868199486169]

# returning the output 

10-element QuVector in FiniteBasis{Orthonormal,1}:
...coefficients: Array{Float64,1}
[0.6065306597126334,0.6065306597126334,0.4288819424803534,0.24761510494160166,0.12380755247080083,0.05536842069051653,0.02260406309258736,0.008543532794657104,0.0030205949871958465,0.001006864995731949]


coherentstatevec(n::Int64,alpha::Number) at QuBase.jl/src/arrays/constructors.jl:28
10-element QuVector in FiniteBasis{Orthonormal,1}:
...coefficients: Array{Float64,1}
[0.6065306597114603,0.6065306597355303,0.4288819421804789,0.24761510797604577,0.12380752738214919,0.05536859516814174,0.02260302507996636,0.008548870306885925,0.0029967233467164565,0.0011000726308558629]
```

src/arrays/constructors.jl
```diff
@@ -16,4 +16,17 @@
     one_at_ind!(arr, i) = setindex!(arr, one(eltype(arr)), i)
     single_coeff(i, lens...) = one_at_ind!(zeros(lens), i)
 
-export statevec
\ No newline at end of file
+function coherentstatevec(n::Int,alpha::Number;method="direct")
+    s = zeros(n)
```
acroy: This is actually more tricky. If you pass a complex `alpha` you will get `InexactError` since `s` is an array of floats. I think using `zeros(typeof(float(alpha)),n)` might work (`float` is necessary because you might input an `Int`, but the coefficients are clearly not integer).

However, I am not so happy with having two functions `coherentstatevec` differing only in the presence of a keyword. Either you merge them into one function and pick the method according to the keyword or you rename this one to `coherentstatevec_inf` or such and don't export it. I vote for the latter, since this function strictly only works for infinite `n` and one might get hard to track errors if one uses it for smallish `n`.

Moreover, please add more comments. I know that we are/have been very sloppy about this, but in particular for the "physics-related" stuff (special operators and states etc) I think we should add information about which formulas we are implementing, possibly with a reference. Already in the near future we will start to scratch our heads and wonder what we actually did a couple of months back.

amitjamadagni: We now have the type dependent zero array.

I have changed the function to `coherentstatevec_inf` and have removed the use of keyword.

I have added references to the implementations. But yeah I still need to add references to the implementations from the previous pulls. I guess we could have a separate PR for that.

acroy:
I solved the issue with the Travis failure (see #26). Please rebase and squash after addressing the final comments.

src/arrays/constructors.jl
```diff
@@ -16,4 +16,22 @@
...
+    return scale!(exp(-abs2(alpha)/2),z)
+end
+
+# Reference :
+# 2nd Defintion : http://en.wikipedia.org/wiki/Coherent_states#Mathematical_features_of_the_canonical_coherent_states
+coherentstatevec(n::Int, alpha::Number) = displaceop(n,alpha)*statevec(1,FiniteBasis(n))
+
+export statevec,
+    coherentstatevec,
+    coherentstatevec_inf
```
acroy: We shouldn't export this function. It probably will cause trouble. Let's just use it internally and if there is demand we can export later.

test/operatortest.jl
```diff
@@ -10,11 +10,24 @@
...
+# Position, Displacement & Momentum Operators Test #
+####################################################
 
 p = positionop(2)
 m = momentumop(2)
 @assert coeffs(commutator(sigma_x, p)) == spzeros(2,2)
 @assert coeffs(commutator(sigma_y, m)) == spzeros(2,2)
+
+csvd = coherentstatevec_inf(20,1)
+csvo = displaceop(20,1)*statevec(1,FiniteBasis(20))
```
acroy: Maybe call `coherentstatevec` instead? It would be more readable and the definition is the same.

src/arrays/constructors.jl
amitjamadagni: But we are using it one of the tests. So I guess we need to export it.

test/operatortest.jl
```diff
@@ -10,11 +10,24 @@
...
+@test_approx_eq_eps coeffs(csvd)[1] coeffs(csvo)[1] 1e-8
+
+############################
+# Squeezing Operators Test #
+############################
+
+@assert squeezingop(2,1.0)== QuArray(eye(2))
+@assert squeezingop(lowerop(2), lowerop(2), 1.0) == squeezingop(2,1)
```
acroy: This is redundant since the left hand side is the definition of `squeezingop`. You could call `@assert squeezingop(lowerop(2), QuArray(eye(2)), 2.0) == displacementop(2,1.0)`, no?

src/arrays/constructors.jl
acroy: You can call it via `QuBase.coherentstatevec_inf`. (In Julia you can always access all methods and types defined in a module by putting `Modulename.` in front. Exported methods are in a way just a convenience and in that sense define a "public API".)

amitjamadagni:
@acroy 
I have tried these things out. Locally they pass for me, but the build-bot shows an error 

```julia
println(coeffs(displaceop(2,1.0)'))
println(coeffs(squeezingop(lowerop(2), QuArray(eye(2)),2.0)))
println(@assert coeffs(squeezingop(lowerop(2), QuArray(eye(2)), 2.0)') == coeffs(displaceop(2,1)))
println(@assert squeezingop(lowerop(2),QuArray(eye(2)), 2.0)' == displaceop(2,1)) 

[0.5403023058681398 0.8414709848078965
 -0.8414709848078965 0.5403023058681397]
[0.5403023058681398 0.8414709848078965
 -0.8414709848078965 0.5403023058681397]
nothing
nothing
```

acroy:
Does it make a difference if you take `coeffs(displaceop(2,1.0))` in the last assert (note the 1.0)? Otherwise you have to resort to `@test_approx_eq` or `@test_approx_eq_eps`.

amitjamadagni:
But locally for me it is passing, I guess it should pass by the build-bot as well, right ? I am missing something here ??

acroy:
It's strange but it can happen. In the worst case you have to put in println statements like in your local test.

acroy:
The `==` in your `@test_approx_eq` commit is superfluous ...

acroy:
Hah, nice! Could you please squash your commits. I will have another look tomorrow, but I think everything looks good now.

PR #26, opened 2015-04-20
Resolve issue with new Tuple syntax in LabelBasis.

acroy:
This is an attempt to resolve the Travis failure for julia 0.4 by using the Compat package and new Tuple syntax.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: julia
+os:
+    - linux
+    - osx
 julia:
     - release
     - nightly
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
-julia 0.3-
+julia 0.3
+Compat 0.4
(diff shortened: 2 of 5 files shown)


acroy:
Has anyone an idea what goes wrong? It's probably something trivial, but I can't figure it out ...

acroy:
Ok, my original problem was somehow related to my editor not inserting the right kind of line break (or me being to stupid). Anyways, Travis works again, but there are tons of warnings. I am not sure if we want to address them now or wait until 0.4 has stabilized? But this should be a separate PR ...

jrevels:
Awesome job, thanks for figuring this out! Sorry about not lending a hand - I was busy with QuDirac stuff yesterday.

acroy:
@jrevels No problem; the hardest part was figuring out the missing new line character :-) What shall we do about the warnings?

PR #27, opened 2015-04-21
Rename (anti)commutator to (anti)commute.

acroy:
IMHO this is more consistent with `scale`, `add` etc. Moreover, I would like to reserve `Commutator` and `Anticommutator` for lazy (anti)commutators or types representing the respective super-operators.

### DIFF
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -129,15 +129,16 @@ end
 
 tensor{B<:OrthonormalBasis}(bra::DualVector{B}, ket::QuVector{B}) = tensor(ket, bra)
 
-##############
-# Commutator #
-##############
+###############
+# Commutators #
+###############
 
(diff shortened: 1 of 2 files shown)


amitjamadagni:
I guess it would be great if we could move the comments from the tests to the main code, as this would help in generating docs and also keep the test code free from docs. In sense we could have some examples as well in the docs and these examples would act as tests, further helping us to understand.

acroy:
Yeah, maybe it is possible to cross-link to the docs in the main code? The comments now were just intended to remind us of the purpose of the tests.

amitjamadagni:
I will trying going through the doc packages if we can do the cross-link. Sure, these things will definitely help us in the future :)

jrevels:
Good call. `Commutator`/`Anticommutator` types would indeed be cool. 

This looks good to me, I'll merge it in.

PR #28, opened 2015-04-24
Spin ops refactor for fewer temporary arrays and type stability

jrevels:
The type instability of `spin` was making me uneasy, so I fixed it by making it an actual value, which I'm now convinced makes more sense in this context. 

In addition, this PR refactors the construction functions we had to avoid making as many temporary arrays.

### DIFF
--- a/src/arrays/spinops.jl
+++ b/src/arrays/spinops.jl
@@ -1,71 +1,80 @@
 #########################
 # Spin Operator Methods #
 #########################
-
-immutable HalfSpin{S}
-    function HalfSpin()
-        S >= 0 ? new() : error("Spin must be positive")
-    end
+immutable HalfSpin{T<:Integer}
+    val::T
+    HalfSpin(val) = val >= 0 ? new(val) : error("Spin must be positive")
 end
 
-# The `spin` function takes the `j` argument from the value
-# domain to the type domain, making the `spin` function type unstable.
+HalfSpin{T}(val::T) = HalfSpin{T}(val)
 
-spin(j::Integer) = HalfSpin{2*j}()
-
-function spin(j::Float64)
+spin(s::HalfSpin) = s
+spin(j::Integer) = HalfSpin(2*j)
+function spin(j)
     if isinteger(j)       
         return spin(int(j)) # j = 0.0, 1.0, 2.0...
     elseif isinteger(2*j) 
-        return HalfSpin{int(2*j)}() # j = .5, 1.5, 2.5...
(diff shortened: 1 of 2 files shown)


acroy:
Thanks, Jarrett! Looks good to me. If you happen to have a reference for the spin operators, it might be good to include it.

jrevels:
You mean just for explanation's sake? I just added some references to the spin and ladder operators wikipedia pages, let me know if that suffices.

Also, why did we named these functions `spin_Jx`, `spin_Jy`, `spin_Jz` instead of `spinx`, `spiny`, `spinz`? The `J` makes me think of orbital angular momentum rather than spin. I see that they use `J` [here](http://en.wikipedia.org/wiki/Ladder_operator#Angular_momentum), though. Is this common/standard? My inexperience is starting to show :)

In the same vein, if we did rename them to `spinx` etc., perhaps we could rename `spin_Jp` to `spinplus` and `spin_Jm` to `spinminus`?

BTW, it seems addition/subtraction wasn't imported, so I fixed that on master and merged it back in.

jrevels:
Also @amitjamadagni my apologies for all of this review/revision happening so late after the initial spin PR, I was quite busy during that time and didn't review as thoroughly as I should have.

amitjamadagni:
@jrevels sure, no problem, please do let me know if I can chip in here and make any contributions. Thanks.

acroy:
> You mean just for explanation's sake? I just added some references to the spin and ladder operators wikipedia pages, let me know if that suffices.

Yes. I think we should start to add such things at least for the more physics-related stuff. We will have to properly document it anyways and then it will be helpful to have some hint where to look it up.

> Also, why did we named these functions spin_Jx, spin_Jy, spin_Jz instead of spinx, spiny, spinz?

Dunno. I am not much of a spin-person, but I guess **J** is rather angular momentum and **S** is spin. [QuTiP](http://qutip.org/docs/2.2.0/apidoc/functions.html#qutip.operators.jmat) uses J for some reason and I think this was reason for Amit's choice. I am fine with both, `spinJ` or renaming everything to just `spin`. As long as we are clear about the definition, it shouldn't really matter.

jrevels:
After referring to the QuTip page, I just went with removing the caps and underscores to make things easier to type (`spinjx`, `spinjy` etc.).

Any qualms before I merge this in (after the build passes, of course)?

acroy:
Thanks! Looks good to me. If you feel like it, maybe squashing into fewer/single commits would be nice.

PR #34, opened 2015-05-14
Added matrix_division support for QuArray.

amitjamadagni:
A*X ==  B where A,X,B are `AbstractQuArray`.

### DIFF
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -27,13 +27,13 @@ end
 # operator * ket -> ket
 function *{B<:OrthonormalBasis}(op::AbstractQuMatrix{B}, ket::AbstractQuVector{B})
     vc = rawcoeffs(op)*rawcoeffs(ket)
-    QAT = similar_type(typeof(ket))
+    QAT = similar_type(ket)
     return QAT(vc, bases(op,1))
 end
 
 function *{B<:OrthonormalBasis}(op::DualMatrix{B}, ket::AbstractQuVector{B})
     vc = Ac_mul_B(rawcoeffs(op), rawcoeffs(ket))
-    QAT = similar_type(typeof(ket))
+    QAT = similar_type(ket)
     return QAT(vc, bases(op,1))
 end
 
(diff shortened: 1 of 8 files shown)


src/arrays/arraymath.jl
```diff
@@ -109,7 +109,7 @@ Base.scale(qarr::AbstractQuArray, num::Number) = scale(num, qarr)
 *(num::Number, qarr::AbstractQuArray) = scale(num, qarr)
 *(qarr::AbstractQuArray, num::Number) = scale(qarr, num)
 /(qarr::AbstractQuArray, num::Number) = scale(1/num, qarr)
-
+\(qarr::AbstractQuArray, v::AbstractQuArray) = QuArray(\(coeffs(qarr), coeffs(v)))
```
acroy: You should determine the return type using `similar_type` (see +,-,* for example). It's probably better to restrict the arguments to 2 `AbstractQuMatrix` and `AbstractQuMatrix` and `AbstractQuVector` (so you need two separate functions). Anything else won't work.

src/arrays/arraymath.jl
```diff
@@ -110,6 +108,17 @@ Base.scale(qarr::AbstractQuArray, num::Number) = scale(num, qarr)
 *(qarr::AbstractQuArray, num::Number) = scale(qarr, num)
 /(qarr::AbstractQuArray, num::Number) = scale(1/num, qarr)
 
+function \{B<:OrthonormalBasis}(qm1::AbstractQuMatrix{B}, qm2::AbstractQuMatrix{B})
+    div = \(rawcoeffs(qm1),rawcoeffs(qm2))
```
acroy: Why not `coeffs`?

acroy:
Please add brief comments stating what the functions are doing (solving linear systems) and some simple tests.

amitjamadagni:
@acroy the tests seem to pass on julia 0.3 and fail on the other julia 0.4,  and also I have hard coded the tests,  any better way around would be really helpful.

test/multest.jl
```diff
@@ -31,3 +31,9 @@ qv = QuArray(v)
 
 # a simple test of the `==` operator
 @assert qm*3im == scale!(3im, copy(qm))
+
+# tests for Matrix Division
+v1 =  [0.5+0*im, 0.+0.*im]
+qv1 = normalize!(QuArray(v1))
+@assert coeffs(\(sigmaz, sigmax)) == sparse([0. 1.;-1. 0.])
```
acroy: Try to use `full(sigmaz)` and `full(sigmax)` here and in the next line. As I said via PM it might actually be better for the Pauli matrices to be dense anyways, but this can be done separately.

acroy:
Since #33 made some changes in arraymath you will need to rebase against master before we are able to merge.

PR #35, opened 2015-05-21
Pauli to dense, matrix division and tests added.

amitjamadagni:
From the #34 converting pauli to dense using full.
Matrix division support added for A*X == B where A & B are`AbstractQuMatrix` & `AbstractQuMatrix` and A & B are `AbstractQuMatrix` and `AbstarctQuMatrix`.
Tests added.


acroy:
Looks good to me.

jrevels:
Looks good to me as well! Ready to merge?

acroy:
Sure, go ahead.

Just as an info: we decided making `sigma*` dense, because the sparse format introduces quite some memory overhead for those small matrices and they are actually not really sparse (2 of 4 elements nonzero).

jrevels:
Makes sense! Hopefully fixed-size matrices will become a thing sometime soon...

acroy:
That would be really awesome!

amitjamadagni:
:+1 that was fast :)

PR #36, opened 2015-05-24
Dot product for QuArray's.

amitjamadagni:
I was going through the material on Krylov propagator, as it requires the dot of QuArray's I have added this, I am not really sure if we require this , any comments on this would be helpful.

### DIFF
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -124,6 +124,21 @@ end
 #  Vectorize QuArray
 Base.vec{B<:OrthonormalBasis}(vec1::AbstractQuArray{B}) = QuArray(vec(coeffs(vec1)))
 
+# Trace of AbstractQuMatrix
+Base.trace{B<:OrthonormalBasis}(qm1::AbstractQuMatrix{B}) = trace(coeffs(qm1))
+
+# dot product of QuArray's
+Base.dot{B<:OrthonormalBasis}(vec1::AbstractQuVector{B}, vec2::AbstractQuVector{B}) = dot(coeffs(vec1), coeffs(vec2))
+Base.dot{B<:OrthonormalBasis}(vec1::DualVector{B}, vec2::DualVector{B}) = dot(vec(coeffs(vec1)), vec(coeffs(vec2)))
+Base.dot{B<:OrthonormalBasis}(vec1::AbstractQuVector{B}, vec2::DualVector{B}) = error("Inner product of a dual vector and a vector is not supported.")
+Base.dot{B<:OrthonormalBasis}(vec1::DualVector{B}, vec2::AbstractQuVector{B}) = error("Inner product of a dual vector and a vector is not supported.")
+
+# Reference : http://en.wikipedia.org/wiki/Dot_product#Dyadics_and_matrices
+Base.dot{B<:OrthonormalBasis}(qm1::AbstractQuMatrix{B}, qm2::AbstractQuMatrix{B}) = trace(qm1'*qm2)
+Base.dot{B<:OrthonormalBasis}(qm1::DualMatrix{B}, qm2::AbstractQuMatrix{B}) = trace(qm1*qm2)
+Base.dot{B<:OrthonormalBasis}(qm1::DualMatrix{B}, qm2::DualMatrix{B}) = trace(qm1*qm2')
+Base.dot{B<:OrthonormalBasis}(qm1::AbstractQuMatrix{B}, qm2::DualMatrix{B}) = trace(qm1'*qm2')
+
 # matrix operations returning a scalar
 # normalization
 Base.norm(qarr::AbstractQuArray) = vecnorm(rawcoeffs(qarr))
--- a/test/multest.jl
+++ b/test/multest.jl
@@ -38,5 +38,18 @@ qv1 = normalize!(QuArray(v1))
 @assert coeffs(\(sigmax,qv1)) == [0.+0.*im, 1.+0.*im]
 @assert coeffs(\(sigmaz, sigmax)) == [0. 1.;-1. 0.]
 
+# Trace
+# Pauli matrices sigmax, sigmay, sigmaz are traceless
+# Ref : http://en.wikipedia.org/wiki/Pauli_matrices#Algebraic_properties
+@assert trace(sigmax) == zero(eltype(sigmax))
+@assert trace(sigmax) == trace(sigmaz)
+
+# Dot product
+# Being defined both for vectors and dual vectors
+# Related Ref : https://github.com/JuliaLang/julia/issues/11064
+@assert dot(qv,qv) == dot(qv',qv')
+# dot(A,B)=trace(A'*B) for the case of sigmax, sigmay results in no trace.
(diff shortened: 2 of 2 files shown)


acroy:
+1 for `dot`. We should also have this for `QuMatrix`. Related JuliaLang/julia#11064.

acroy:
The Travis error for 0.4 seems unrelated, but we should have a closer look.

src/arrays/arraymath.jl
```diff
@@ -121,6 +121,11 @@ function \{B<:OrthonormalBasis}(op::AbstractQuMatrix{B}, vec::AbstractQuVector{B
     return QAT(div, bases(op,1))
 end
 
+# dot product of QuArray's
+dot{B<:OrthonormalBasis}(vec1::AbstractQuVector{B}, vec2::AbstractQuVector{B}) = dot(coeffs(vec1), coeffs(vec2))
+# Reference : http://en.wikipedia.org/wiki/Dot_product#Dyadics_and_matrices
+dot{B<:OrthonormalBasis}(qm1::AbstractQuMatrix{B}, qm2::AbstractQuMatrix{B}) = trace(coeffs(qm1')*coeffs(qm2))
```
acroy: I think it is better to use `trace(coeffs(qm1'*qm2))` here. If you want you can also steal `trace` for `QuMatrix` from #32. We have to check the issue I linked to if there is a more efficient way to do this dot product.

src/arrays/arraymath.jl
```diff
@@ -121,6 +121,11 @@ function \{B<:OrthonormalBasis}(op::AbstractQuMatrix{B}, vec::AbstractQuVector{B
     return QAT(div, bases(op,1))
 end
 
+# dot product of QuArray's
+dot{B<:OrthonormalBasis}(vec1::AbstractQuVector{B}, vec2::AbstractQuVector{B}) = dot(coeffs(vec1), coeffs(vec2))
```
acroy: Do we need extra definitions for the cases where one vector is actually a dual vector in which case we should throw an error?

jrevels: +1, we shouldn't allow `DualVector`s as arguments to `dot` - Julia's convention for `dot` is that it takes in two column vectors and applies the transpose/conjugate transpose as necessary.

acroy: But we can allow two `DualVectors`, right?

jrevels: Hmm...I don't see why not. Currently `dot` throws an error if you try to input row vectors, but I suppose that's due more to the matrix/row vector type ambiguity rather than an actual API decision.

amitjamadagni:
@acroy @jrevels I have made an attempt at dot (considering dual vectors as well) and trace, added tests for the same. A review would be helpful.

test/multest.jl
```diff
@@ -37,3 +37,8 @@ v1 =  [0.5+0*im, 0.+0.*im]
 qv1 = normalize!(QuArray(v1))
 @assert coeffs(\(sigmax,qv1)) == [0.+0.*im, 1.+0.*im]
 @assert coeffs(\(sigmaz, sigmax)) == [0. 1.;-1. 0.]
+
+# tests for trace, dot
```
acroy: I would prefer the tests being easier to understand on a first glance, which implies more comments. If one doesn't know about Pauli matrices it is unclear what the tests are doing. I think it is better to say for instance
```
# Pauli matrices sigmax, sigmay, sigmaz are traceless (see ... )
@assert trace(sigmax) == zero(eltype(sigmax))
@assert trace(sigmaz) == zero(eltype(sigmaz))
```
and so on.

acroy:
Apart from the comment on the tests, I think this looks good.

amitjamadagni:
@acroy edits done. If this is fine, I will squash the commits ?

test/multest.jl
```diff
@@ -37,3 +37,15 @@ v1 =  [0.5+0*im, 0.+0.*im]
...
+# Pauli matrices sigmax, sigmay, sigmaz are traceless
+# Ref : http://en.wikipedia.org/wiki/Pauli_matrices#Algebraic_properties
+@assert trace(sigmax) == zero(eltype(sigmax))
+@assert trace(sigmax) == trace(sigmaz)
+
+# Dot product
+# Being defined both for vectors and dual vectors
+# Related Ref : https://github.com/JuliaLang/julia/issues/11064
+@assert dot(qv,qv) == dot(qv',qv')
+@assert dot(sigmax, sigmay) == trace(sigmaz)
```
acroy: This line also needs an explanation ...

amitjamadagni:
@acroy done!

acroy:
Good. Please squash the commits and then we can merge I would say.

acroy:
Sorry. I just noticed that you squashed already :-)

amitjamadagni:
@acroy can we merge this ?

acroy:
Basically yes. I was just thinking if we should generalize dot to allow for two `AbstractQuVector`s (and giving an error in the mixed case)?

amitjamadagni:
@acroy I have tried creating a common `dot` for `AbstractQuVector`.

acroy:
I see, but what I meant was the following
```
dot{B<:OrthonormalBasis}(vec1::AbstractQuVector{B}, vec2::AbstractQuVector{B}) = ...
dot{B<:OrthonormalBasis}(vec1::DualVector{B}, vec2::DualVector{B}) = ...
dot{B<:OrthonormalBasis}(vec1::DualVector{B}, vec2::AbstractQuVector{B}) = error("Inner product of a dual vector and a vector is not supported.")
dot{B<:OrthonormalBasis}(vec1::AbstractQuVector{B}, vec2::DualVector{B}) = error("Inner product of a dual vector and a vector is not supported.")
```
and maybe similar for `AbstractQuMatrix`.

amitjamadagni:
@acroy the latest commit covers the cases, please do let me know if I am missing something here.

acroy:
You are checking if the element types of the two vectors are the same, while I simply want to prevent a `DualVector` and an `AbstractQuVector` as parameters.

amitjamadagni:
Sorry my bad I understand what I did here. So what would be the best way out, the initial version or the latest version ?

acroy:
Either you revert it to your original version and we sort it out later or you just take the code I posted above and replace the `...` with the code you provided originally.

amitjamadagni:
Would be good if we compare the types of vectors i.e., `typeof(vec1) == typeof(vec2)` which would then generalize to comparing `QuArray` and `CTranspose` ?

acroy:
You don't need to compare the types. `dot{B<:OrthonormalBasis}(vec1::AbstractQuVector{B}, vec2::AbstractQuVector{B})` actually applies to all vectors (including dual vectors). Now we have to provide more specific versions to exclude the cases with mixed parameters. All the work will be done by the compiler.

amitjamadagni:
@acroy a review would be helpful.

acroy:
@amitjamadagni : Could you please squash and rebase?

amitjamadagni:
@acroy  done !

acroy:
@amitjamadagni : Thanks!

PR #37, opened 2015-06-24
Vectorize and tests added.

amitjamadagni:


acroy:
Seems some of the commits for `dot` and `trace` got mixed in.

You should also use `similar_type` vec to get the right constructor (see for instance `/`).

acroy:
Also: does `vec` work for arbitrary arrays or only for matrices?

amitjamadagni:
@acroy `vec` works for arbitrary arrays, I was of the thinking that if it is a vector it returns a vector, if it is a matrix then it flattens it, this might be directly useful in `propagate` methods for example [here](https://gist.github.com/amitjamadagni/4aeeeeca18610cf00e1f), for current_qustate and always returns a `QuArray` as it is always a one dimensional array.

acroy:
Ok, I wasn't sure if it would work for all arrays. You also right about returning `QuArray`.

PR #38, opened 2015-07-12
Complex coeffs support.

amitjamadagni:
Currently we get an error doing this : 

``` julia
julia> statevec(1, FiniteBasis(2))
2-element QuVector in FiniteBasis{Orthonormal}:
...coefficients: Array{Float64,1}
[1.0,0.0]

julia> complex(statevec(1, FiniteBasis(2)))
ERROR: `complex` has no method matching complex(::QuArray{FiniteBasis{Orthonormal},Float64,1,Array{Float64,1}})
```

This further gives an error when we do something like this : 

``` julia
julia> using QuDynamics

julia> qode45 = QuPropagator(sigmax, statevec(1, FiniteBasis(2)), 0.:0.1:2*pi, QuODE45())
QuPropagator{QuODE45,QuArray{FiniteBasis{Orthonormal},Float64,1,Array{Float64,1}},QuSchrodingerEq{H<:AbstractQuArray{B<:AbstractBasis{S<:AbstractStructure},T,2}}}(QuSchrodingerEq{QuArray{FiniteBasis{Orthonormal},Float64,2,Array{Float64,2}}}(2x2 QuMatrix in FiniteBasis{Orthonormal}:
...coefficients: Array{Float64,2}
[0.0 1.0
 1.0 0.0]),2-element QuVector in FiniteBasis{Orthonormal}:
...coefficients: Array{Float64,1}
[1.0,0.0],0.0:0.1:6.2,QuODE45(Dict{Symbol,Any}()))

julia> next_qode45 = next(qode45, start(qode45))
ERROR: InexactError()
 in setindex! at array.jl:307
 in copy! at abstractarray.jl:149
 in setindex! at array.jl:307
 in oderk_adapt at .julia/v0.3/ODE/src/runge_kutta.jl:279
 in ode45_dp at .julia/v0.3/ODE/src/runge_kutta.jl:212
 in propagate at .julia/v0.3/QuDynamics/src/propodesolvers.jl:28
 in next at .julia/v0.3/QuDynamics/src/propmachinery.jl:47
```

This PR aims to fix this.

### DIFF
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -122,7 +122,7 @@ function \{B<:OrthonormalBasis}(op::AbstractQuMatrix{B}, vec::AbstractQuVector{B
 end
 
 # convert coeffs to complex
-Base.complex{B<:OrthonormalBasis}(qarr::AbstractQuArray{B}) = similar_type(qarr)(complex(coeffs(vec(qarr))), bases(qarr))
+Base.complex{B<:OrthonormalBasis}(qarr::AbstractQuArray{B}) = similar_type(qarr)(complex(coeffs(qarr)), bases(qarr))
 
 #  Vectorize QuArray
 Base.vec{B<:OrthonormalBasis}(vec1::AbstractQuArray{B}) = QuArray(vec(coeffs(vec1)))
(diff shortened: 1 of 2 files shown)


acroy:
Definitely required. There are, however, more functions like `complex`. So the question is if/how we can support all of them in an elegant way?

amitjamadagni:
@acroy could we have an issue for all the functions or we could just list them out, there was a use case so I thought of this but if we could list all at one place we would get a more neater picture.

acroy:
The "easy" way would be to use code generation to provide functions for `float`, `float16`, `float32`, etc. But I have the feeling that this is not the best (and future proof) way to do that. Maybe we just need a bunch of `convert`s. Anyways I think it would be good to look at other Array-like types and figure out how this is handled.

PR #39, opened 2015-07-16
Multiplication fix.

amitjamadagni:
The following case fails 

``` julia
julia> a = full(tensor(lowerop(2), QuArray(eye(2))))
4x4 QuMatrix in FiniteBasis{Orthonormal}:
...coefficients: Array{Float64,2}
[0.0 0.0 1.0 0.0
 0.0 0.0 0.0 1.0
 0.0 0.0 0.0 0.0
 0.0 0.0 0.0 0.0]

julia> sm = full(tensor(QuArray(eye(2)), lowerop(2)))
4x4 QuMatrix in FiniteBasis{Orthonormal}:
...coefficients: Array{Float64,2}
[0.0 1.0 0.0 0.0
 0.0 0.0 0.0 0.0
 0.0 0.0 0.0 1.0
 0.0 0.0 0.0 0.0]

julia> a*sm'
4x4 QuMatrix in FiniteBasis{Orthonormal}:
...coefficients: Array{Float64,2}
[0.0 0.0 0.0 0.0
 0.0 0.0 0.0 0.0
 0.0 1.0 0.0 0.0
 0.0 0.0 0.0 0.0]
```

the last one has to be 

``` julia
julia> a*sm'
4x4 QuMatrix in FiniteBasis{Orthonormal}:
...coefficients: Array{Float64,2}
[0.0 0.0 0.0 0.0
 0.0 0.0 1.0 0.0
 0.0 0.0 0.0 0.0
 0.0 0.0 0.0 0.0]
```

### DIFF
--- a/test/multest.jl
+++ b/test/multest.jl
@@ -7,27 +7,29 @@ v = [1.0,2.0,3.0]
 qm = QuArray(m)
 qv = QuArray(v)
 
-@assert rawcoeffs(qm*qm) == m*m
+@assert coeffs(qm*qm) == m*m
+@assert coeffs(qm'*qm) == m'*m
+@assert coeffs(qm*qm') == m*m'
+@assert coeffs(qm'*qm') == m'*m'
 @assert rawcoeffs(qm'*qm') == m*m
-@assert rawcoeffs(qm'*qm) == m'*m
-@assert rawcoeffs(qm*qm') == m*m'
 
-@assert rawcoeffs(qm*qv) == m*v
+@assert coeffs(qm*qv) == m*v
 @assert rawcoeffs(qv'*qm') == m*v
-@assert rawcoeffs(qm'*qv) == m'*v
-@assert rawcoeffs(qv'*qm) == m'*v
+@assert coeffs(qv'*qm') == v'*m'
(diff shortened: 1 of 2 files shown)


acroy:
Well spotted! Could you add a test for this case? (I guess something like `sigmax*sigmaz'` vs `coeffs(sigmax)*coeffs(sigmaz)'` would be sufficient)

amitjamadagni:
@acroy tests added and squashed. Ready to be merged !

test/multest.jl
```diff
@@ -11,6 +11,7 @@ qv = QuArray(v)
 @assert rawcoeffs(qm'*qm') == m*m
```
acroy: Uh, I guess all of those should be `coeffs` rather than `rawcoeffs`? This would probably also cover the new test case?

amitjamadagni: `rawcoeffs` and `coeffs` return the same for all the cases except for the recent commit. @acroy please do let me know if am missing something further here.

acroy: You are right, although we probably could/should consistently use `coeffs` everywhere (unless we really mean `rawcoeffs`).

amitjamadagni:
@acroy I hope this is ready.

PR #40, opened 2015-07-31
Expectation added.

amitjamadagni:
Calculating the expectation for `AbstractQuVector` and `AbstractQuMatrix`.

### DIFF
--- a/src/arrays/arraymath.jl
+++ b/src/arrays/arraymath.jl
@@ -198,8 +198,15 @@ tensor(bra::DualVector, ket::AbstractQuVector) = tensor(ket, bra)
 commute(a::AbstractQuMatrix, b::AbstractQuMatrix) = (a*b) - (b*a)
 anticommute(a::AbstractQuMatrix, b::AbstractQuMatrix) = (a*b) + (b*a)
 
+################
+#  Expectation #
+################
+expectationvalue(qarr::AbstractQuVector, op::AbstractQuMatrix) = qarr'*op*qarr
+expectationvalue(qarr::AbstractQuMatrix, op::AbstractQuMatrix) = trace(qarr*op)
+
 export normalize,
     normalize!,
     tensor,
     commute,
-    anticommute
+    anticommute,
+    expectationvalue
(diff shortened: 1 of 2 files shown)


acroy:
The question is if we want to restrict one of the arguments to explicitly be a state (vector) - so we would need something like #32?

acroy:
I think this is actually Ok for now. I would vote for changing the function name to something like `expectationvalue` (is there something similar in QuDirac?). Tests are missing though.

amitjamadagni:
@acroy I guess this is ready ! The tests pass on the release version !

amitjamadagni:
Also a quick search on QuDirac didnt give any results for something similar : 

```
amit@esornep:~/Downloads/QuDirac.jl$ cd src/
amit@esornep:~/Downloads/QuDirac.jl/src$ grep -r -i "expect" .
amit@esornep:~/Downloads/QuDirac.jl/src$ grep -r -i "expectation" .
amit@esornep:~/Downloads/QuDirac.jl/src$ grep -r -i "expectationvalue" .

amit@esornep:~/Downloads/QuDirac.jl$ grep -r -i "expectation" .
./docs/op_math.md:Thus, expectation values are naturally obtained in this manner:
amit@esornep:~/Downloads/QuDirac.jl$ grep -r -i "expect" .
./docs/op_math.md:Thus, expectation values are naturally obtained in this manner:
./docs/d_str.md:When using `d"..."` literals, the `|`, `>` and `<` characters *cannot be used for anything other than as symbols for Kets, Bras, and inner products*. Other than that, any other Julia syntax should evaluate properly. For example, assignments and function calls work as expected:
amit@esornep:~/Downloads/QuDirac.jl$ grep -r -i "expectionvalue" .
```

src/arrays/arraymath.jl
```diff
@@ -198,8 +198,15 @@ tensor(bra::DualVector, ket::AbstractQuVector) = tensor(ket, bra)
 commute(a::AbstractQuMatrix, b::AbstractQuMatrix) = (a*b) - (b*a)
 anticommute(a::AbstractQuMatrix, b::AbstractQuMatrix) = (a*b) + (b*a)
 
+###############
+#  Expectaion #
```
acroy: typo: "Expectation"

amitjamadagni: Fixed !

PR #44, opened 2015-10-02
Sparse support for AbstractQuMatrix.

amitjamadagni:
Sparse support added for `AbstractQuMatrix`. Use of `coeffs` instead of `rawcoeffs`, in `full` and `expm`.


amitjamadagni:
@acroy I guess this is ready !

PR #47, opened 2016-08-22
Normalize using Compat.

amitjamadagni:


amitjamadagni:
@acroy `normalize` is breaking other packages (for example https://github.com/JuliaQuantum/QuDynamics.jl/pull/42), it would be helpful if we could have a quick review on this. Thanks !

src/arrays/arraymath.jl
```diff
@@ -143,12 +143,12 @@ Base.dot{B<:OrthonormalBasis}(qm1::AbstractQuMatrix{B}, qm2::DualMatrix{B}) = tr
 # normalization
 Base.norm(qarr::AbstractQuArray) = vecnorm(rawcoeffs(qarr))
 
-function normalize!(qarr::AbstractQuArray)
+function Compat.normalize!(qarr::AbstractQuArray)
```
acroy: Does Compat define `normalize!` as well?

amitjamadagni: Yup it is mentioned in the README of the package https://github.com/JuliaLang/Compat.jl

acroy: Ok. Then this should be fine ...

